PR #5, opened 2020-11-13
Update Lhotse's augmentation API usage

pzelasko:
This is to complement the changes that we made in https://github.com/lhotse-speech/lhotse/pull/124 

If you'd like we can also switch to torchaudio augmentation, roughly the same but drops the WavAugment dependency.


danpovey (commented):
Not an expert on these kinds of things, but looks OK to me...

danpovey:
resolving this separately

PR #7, opened 2020-11-13
fix some bugs/issues of the training scripts

qindazhu:


egs/librispeech/asr/simple_v1/prepare.py
```diff
@@ -44,17 +46,7 @@
             supervisions=manifests['supervisions']).compute_and_store_features(
```
qindazhu, line 46: There is a warning here with the latest lhotse
```
UserWarning: ProcessPoolExecutor using a "fork" multiprocessing detected. In some circumstances this can crash or hang the program, e.g. when using  data augmentation libsox wrappers like WavAugment or torchaudio. We suggest passing an extra argument to initialize the executor: ProcessPoolExecutor(..., mp_context=multiprocessing.get_context("spawn"))
  warnings.warn('ProcessPoolExecutor using a "fork" multiprocessing detected. '
```
I believe it can be fixed with @pzelasko's PR https://github.com/k2-fsa/snowfall/pull/5/files

danpovey:
Thanks!  Merging.

qindazhu:
The scripts can now fully run with latest k2 and lhotse, but still have some issues:

1. https://github.com/lhotse-speech/lhotse/issues/125 causes we cannot run many batches successfully.
2. The modle output is diverging, may because learning rate or model struct, we definitely need to look into this.
3. Sometimes we may get `Cuda out of memory`, but because 1, we can not repro it.

PR #9, opened 2020-11-13
revert Piotr's changes

qindazhu:


qindazhu:
Revert it for now, because of https://github.com/lhotse-speech/lhotse/issues/126.

PR #11, opened 2020-11-15
Add non-augmented data in training

pzelasko:
I think this might've been an omission rather than an intention to leave out the non-augmented data in the training. This augmentation setup might still not be perfect, but it's a step closer to what we would have typically done.


qindazhu:
Yeah, back to that time, I did't know well how to do data preparetion with lhotse, I was just copying Junbo's scripts.

PR #33, opened 2020-12-04
fix bug in arpa2fst.py which did not remove </s>

qindazhu:


jimbozhang:
For the length limit of the code line, I think using 120 characters is clearer than 80 characters...

csukuangfj:
> For the length limit of the code line, I think using 120 characters is clearer than 80 characters...


I agree. But the limit of 80 characters encourages us to write more compact code.
The C++ code in k2 uses 80 characters as well.

qindazhu:
I'm OK with both.. anyway, IIRC, we have not set up any style config file in snowfall for now, we can do this later automatically.

egs/librispeech/asr/simple_v1/local/arpa2fst.py
```diff
@@ -51,22 +64,28 @@ class Ngram(NamedTuple):
 def parse_arpa_file(filename, to_natural_log=True):
     ngrams = []
     with open(filename) as f:
-        stage = 0   # stage 0 for header, 1 for unigram, 2 for bigram, ...
+        stage = 0  # stage 0 for header, 1 for unigram, 2 for bigram, ...
         for line in f:
             line = re.sub(r'\s+$', '', line)
-            if re.match(r'^\s*$', line) or re.match(r'\\(data|end)\\', line) or re.match(r'ngram \d+=', line):
+            if re.match(r'^\s*$', line) or re.match(
```
csukuangfj, line 70: I would suggest using `re.compile` to create an object and using the object for matching,
which is more efficient when the ARPA file contains lots of lines.

egs/librispeech/asr/simple_v1/local/arpa2fst.py
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
```
csukuangfj, line 1: Lack of documentation and type hints for functions in this file.

egs/librispeech/asr/simple_v1/local/arpa2fst.py
qindazhu: @jimbozhang it would be appreciated if you can fix these issue, many thanks

jimbozhang: OK. I'll do it later. It is said that Perl programers are not used to compile the regular expressions. 😄

Issue #31, opened 2020-11-27
"Check failed" after decoding + shortest_path


pzelasko:
I've been playing around with the recipe to get to know k2 better. I trained the model for 10 epochs and then tried to perform decoding with the following code (running everything on the CPU):

```python

def create_G_decoding_graph(G, L):
    G = k2.arc_sort(G)
    decoding_graph = k2.intersect(G, L).invert_()
    decoding_graph = k2.add_epsilon_self_loops(decoding_graph)
    return decoding_graph

LG = create_G_decoding_graph(G, L)
LG.scores.requires_grad_(False)

acoustic_fsa = k2.DenseFsaVec(nnet_output, supervision_segments)

lattices = k2.intersect_dense_pruned(LG, acoustic_fsa, search_beam=5.0, output_beam=5.0, min_active_states=10, max_active_states=200)

best_paths = k2.fsa_algo.shortest_path(lattices, True)

# Trying to get the recognized words as strings for the first supervision segment in the batch
hyp_words = [symbol_table._id2sym[x.item()] for x in best_paths[0].aux_labels if x > 0]
```

This crashes with the following message + stack trace:

```
[F] /home/pzelasko/k2-repo/k2/csrc/array.h:Range:107 Check failed: size <= Dim() - start (32525 vs. 0)


[ Stack-Trace: ]
/home/pzelasko/k2-repo/build/lib/libk2_log.so(k2::internal::GetStackTrace()+0x38) [0x7f0cdcef8508]
/home/pzelasko/k2-repo/build/lib/libk2context.so(+0x6de0a) [0x7f0cdd437e0a]
/home/pzelasko/k2-repo/build/lib/libk2context.so(k2::Array1<int>::Range(int, int) const+0x2c7) [0x7f0cdd44ac27]
/home/pzelasko/k2-repo/build/lib/libk2context.so(k2::RaggedShape::Index(int, int, int*)+0x322) [0x7f0cdd50c5f2]
/home/pzelasko/miniconda3/envs/k2env/lib/python3.7/site-packages/_k2.cpython-37m-x86_64-linux-gnu.so(+0x673fc) [0x7f0ce24fc3fc]
/home/pzelasko/miniconda3/envs/k2env/lib/python3.7/site-packages/_k2.cpython-37m-x86_64-linux-gnu.so(+0x17c76) [0x7f0ce24acc76]
[...]

PR #34, opened 2020-12-07
init decode code.

qindazhu:
Caution: the code can not be run now because of issue https://github.com/k2-fsa/snowfall/issues/31#issuecomment-738570169, we are still debugging it, just commit it now so that other guys can work on it together.


qindazhu:
@danpovey you may need to run `pip install editdistance` to install dependency `editdistance`.

egs/librispeech/asr/simple_v1/decode.py
```diff
@@ -0,0 +1,175 @@
...
+    results = decode(dataloader=test_dl,
+                     model=model,
+                     subsampling=subsampling,
+                     device=device,
+                     LG=LG,
+                     symbols=symbol_table)
+    for ref, hyp in results:
+        print('ref=', ref, ', hyp=', hyp)
+    # compute WER
+    dist = sum(editdistance.eval(ref, hyp) for ref, hyp in results)
```
pzelasko, line 163: Use “pip install kaldialign” for identical results with Kaldi

danpovey:
OK cool.  Did you ever get an example of a pair of word symbols that were on the same determinized arc?

danpovey:
.. wonder if it might have been that previous problem with the EOS symbol.

egs/librispeech/asr/simple_v1/decode.py
danpovey: Does it have a similar interface?  Would have to be imported as `import kaldialign`, right?

pzelasko: Yeah. See usage here https://pypi.org/project/kaldialign/

qindazhu:
> OK cool. Did you ever get an example of a pair of word symbols that were on the same determinized arc?

Sorry, I have not debugged into it, if it's urgent, I'll do tomorrow.

danpovey:
Thanks, Piotr!   If you're familiar with the `editdistance` package let us
know what the differences are.
Haowen: not urgent.

egs/librispeech/asr/simple_v1/decode.py
pzelasko: @danpovey I am not very much familiar with `editdistance`. I only vaguely remember evaluating several libs in the past and none of them got me the same WER/alignment as Kaldi, so I ported Kaldi's code to Python myself. A pleasant side effect of `kaldialign` is that the code is way faster (I compared to `Biopython` at that time, but `editdistance` says that it's a pure Python implementation so it's likely slow too). If you want we can just compare the resulting numbers for both libs.

egs/librispeech/asr/simple_v1/decode.py
```diff
@@ -0,0 +1,175 @@
...
+    device_id = 1
+    device = torch.device('cuda', device_id)
+    model = Model(num_features=40, num_classes=364)
+    checkpoint = os.path.join(exp_dir, 'epoch-9.pt')
+    load_checkpoint(checkpoint, model)
+    model.to(device)
+    model.eval()
+
+    LG = LG.to(device)
+    LG.scores.requires_grad_(False)
```
pzelasko, line 152: A question out of curiosity: if we did `requires_grad_(True)` instead, would that have optimized the graph scores (i.e. the product of LM scores and pronunciation scores) alongside the acoustic model?

egs/librispeech/asr/simple_v1/decode.py
```diff
@@ -0,0 +1,175 @@
...
+    # wrong in determinize, we are still debugging the code
+    return
+    # load L, G, symbol_table
+    lang_dir = 'data/lang_nosp'
+    symbol_table = k2.SymbolTable.from_file(lang_dir + '/words.txt')
+
+    if not os.path.exists(lang_dir + '/LG.pt'):
+        print("Loading L_disambig.fst.txt")
+        with open(lang_dir + '/L_disambig.fst.txt') as f:
+            L = k2.Fsa.from_openfst(f.read(), acceptor=False)
```
pzelasko, line 91: Could we have comments stating what's the input and output label when the FSTs are loaded (e.g. "L input symbols: phones; output symbols: words) and when they are inverted? I think it'd be easier to follow.

egs/librispeech/asr/simple_v1/decode.py
```diff
@@ -0,0 +1,175 @@
...
+        print("Loading G.fsa.txt")
+        with open(lang_dir + '/G.fsa.txt') as f:
+            G = k2.Fsa.from_openfst(f.read(), acceptor=True)
+        L = k2.arc_sort(L.invert_())
+        G = k2.arc_sort(G)
+        print("Intersecting L and G")
+        LG = k2.intersect(L, G)
+        print(LG.shape)
+        print("Connect L*G")
+        LG = k2.connect(LG.invert_())
```
pzelasko, line 101: why not connect before inverting? I assume the non-(co)accessible states come from the intersection rather than inversion

egs/librispeech/asr/simple_v1/decode.py
danpovey: OK.  Using kaldialign is fine with me, sounds good.

egs/librispeech/asr/simple_v1/decode.py
danpovey: Mm, I doubt it would have because they are not part of the model.

egs/librispeech/asr/simple_v1/decode.py
danpovey: Good idea, will try to add it at some point.  But remember, in this framework it's not really input and output symbols: it's symbols and aux_symbols.  I'd rather be consistent on this point.

egs/librispeech/asr/simple_v1/decode.py
danpovey: should be equivalent, I believe.

csukuangfj: Even if it is equivalent, the fsa after `k2.connect` will contain fewer number of states and arcs, so `invert_` 
will process fewer data.

egs/librispeech/asr/simple_v1/decode.py
qindazhu: `editdistance` is implemented with C++, there's a benchmark here https://github.com/roy-ht/editdistance.
 Anyway, Just mention it. I'm OK with kaldialing as well.

pzelasko: Oooh I looked up the wrong editdistance (there seem to be more than one libs with that name). Let's use whatever turns out to best.

egs/librispeech/asr/simple_v1/decode.py
pzelasko: It would be sufficient to register `LG.parameters()` in the initialization of the optimizer. Assuming `Fsa` has a method called `parameters()` like `nn.Module`.

egs/librispeech/asr/simple_v1/decode.py
pzelasko: Got it. In that case, it would be super useful if you had a documentation page that presents a mapping between OpenFST concepts and K2 concepts (maybe you have it already).

pzelasko:
I applied the suggestions from the discussion here in #35, we can close this if we merge the other one.

PR #35, opened 2020-12-08
Moving code around + caching training graphs

pzelasko:
I attempted a bit of refactoring, i.e. I created a "library code" directory called "snowfall" so that we can do `from snowfall import X`. Let me know what you think, if you guys like this I can do more work in this direction. I'd like to start building a multilingual recipe using Babel and GlobalPhone sometime soon in the future so I'm trying to identify the bits that will be common/re-used.

I created my branch off of Haowen's "decoding" PR branch so please ignore the first two files in the changelist when reading (I didn't touch them).

I've also attempted caching the training graph compilation since I saw a TODO - however that one raises an error. The only real change is that I'm compiling graphs separately for each phrase, and then creating FsaVec out of a list of Fsas and running `k2.union` on them. The error looks like this:

```
[F] /home/pzelasko/k2-repo/k2/csrc/intersect.cu:k2::MultiGraphDenseIntersect::MultiGraphDenseIntersect(k2::FsaVec&, k2::DenseFsaVec&, float):108 Check failed: a_fsas_.Dim0() == b_fsas_.shape.Dim0()


[ Stack-Trace: ]
/home/pzelasko/k2-repo/build/lib/libk2_log.so(k2::internal::GetStackTrace()+0x38) [0x2aab2e668508]
/home/pzelasko/k2-repo/build/lib/libk2context.so(+0x7884e) [0x2aab2bd9b84e]
/home/pzelasko/k2-repo/build/lib/libk2context.so(k2::MultiGraphDenseIntersect::MultiGraphDenseIntersect(k2::Ragged<k2::Arc>&, k2::DenseFsaVec&, float)+0x923) [0x2aab2bdabc43]
/home/pzelasko/k2-repo/build/lib/libk2context.so(k2::IntersectDense(k2::Ragged<k2::Arc>&, k2::DenseFsaVec&, float, k2::Ragged<k2::Arc>*, k2::Array1<int>*, k2::Array1<int>*)+0x6f) [0x2aab2bd9bc8f]
```

Unfortunately, I built K2 in `Release` so I don't have debug symbols right now; I'll recompile next time. Maybe you'd have an idea what went wrong though.


danpovey:
Great!
Regarding the training graphs compilation: Union() is not the right call.
At the Python level you would want either Stack(axis=0..), if the input had 2 axes (likely) or Append(axis=0..) if the input  had 3 axes.  @csukuangfj and @qindazhu are more familiar with the Python interfaces; do we have anything to combine Fsa's like this?

pzelasko:
Mhm, makes sense, the graphs have to be separate.

qindazhu:
Just remove `Union` would be fine? I thought the insersect result is already `FsaVec`, and seems you forget to cache the result (decoding graph for the texts) with the key (i.e. `texts`)?

pzelasko:
The caching is done [via `lru_cache` decorator](https://docs.python.org/3.7/library/functools.html?highlight=lru_cache#functools.lru_cache), we could set `maxsize=None` to make the cache unbounded (currently it's 50k least recently used elements)

qindazhu:
Oh, I see, you are tryting to cache the graph for each text, then we really need to create a FsaVec from a list of Fsas.

csukuangfj:
https://github.com/k2-fsa/snowfall/blob/7398dcc0177bcf1cb50220e76737c0d44f41db8e/snowfall/training/graph.py#L16

> Just remove Union would be fine? 

Yes, remove `k2.union` should work.

pzelasko:
I confirm that it works with the union removed.

danpovey:
Great!  I am adding you as project admin, so you can merge when you think it's ready.

snowfall/training/graph.py
```diff
@@ -0,0 +1,26 @@
...
+
+import k2
+from k2 import Fsa, SymbolTable
+
+
+class TrainingGraphCompiler:
+    def __init__(self, L: Fsa, vocab: SymbolTable, oov: str = '<UNK>'):
+        self.L = L
+        self.vocab = vocab
+        self.oov = oov
```
csukuangfj, line 12: We need to check that `oov` is inside `self.vocab`.

snowfall/training/graph.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def compile(self, texts: Iterable[str]) -> Fsa:
+        decoding_graphs = k2.create_fsa_vec([self.compile_one_and_cache(text) for text in texts])
+        decoding_graphs.scores.requires_grad_(False)
```
csukuangfj, line 16: I would suggest 
```python
decoding_graphs.requires_grad_(False)
```

It's simpler or just remove this statement since it requires no grad by default.

snowfall/training/graph.py
```diff
@@ -0,0 +1,26 @@
...
+        return decoding_graphs
+
+    @lru_cache(maxsize=50000)
+    def compile_one_and_cache(self, text: str) -> Fsa:
```
csukuangfj, line 20: It may be more efficient to construct a list of linear fsas with a single call of `k2.linear_fsa`.
Not sure when the cache is in scope.

https://github.com/k2-fsa/k2/blob/66da100fc013cbd9a4d7f6ffaf6de3601c6e9ee8/k2/python/k2/fsa_algo.py#L17

```python
def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
    '''Construct an linear FSA from symbols.
    Note:
      The scores of arcs in the returned FSA are all 0.
    Args:
      symbols:
        A list of integers or a list of list of integers.
    Returns:
      An FSA if the input is a list of integers.
      A vector of FSAs if the input is a list of list of integers.
    '''
```

snowfall/training/graph.py
```diff
@@ -0,0 +1,26 @@
...
+        tokens = (token if token in self.vocab._sym2id else self.oov for token in text.split(' '))
+        word_ids = [self.vocab.get(token) for token in tokens]
+        fsa = k2.linear_fsa(word_ids)
+        decoding_graph = k2.intersect(fsa, self.L).invert_()
```
csukuangfj, line 24: I think we need to call `k2.connect` after `k2.intersect` but before `invert_`.

egs/librispeech/asr/simple_v1/train.py
```diff
@@ -64,27 +44,33 @@ def get_tot_objf_and_num_frames(tot_scores, frames_per_seq):
...
     supervisions = batch['supervisions']
     supervision_segments = torch.stack(
         (supervisions['sequence_idx'],
-         torch.floor_divide(supervisions['start_frame'], subsampling),
-         torch.floor_divide(supervisions['num_frames'], subsampling)), 1).to(torch.int32)
-    _,indices = torch.sort(-supervision_segments[:,2])
-    supervision_segments = supervision_segments[indices,:]
+         torch.floor_divide(supervisions['start_frame'], model.subsampling_factor),
+         torch.floor_divide(supervisions['num_frames'], model.subsampling_factor)), 1).to(torch.int32)
+    _, indices = torch.sort(-supervision_segments[:, 2])
```
csukuangfj, line 72: I would suggest `torch.argsort` with `descending=True`:
- It returns only the sorted indices, so it does not require to allocate extra memory for the returned tensor.
- It does not require to negate the input.

egs/librispeech/asr/simple_v1/train.py
```diff
@@ -64,27 +44,33 @@ def get_tot_objf_and_num_frames(tot_scores, frames_per_seq):
...
+    supervision_segments = supervision_segments[indices, :]
```
csukuangfj, line 73: I think just `supervision_segments[indices]` is fine.

snowfall/training/graph.py
```diff
@@ -0,0 +1,26 @@
+from functools import lru_cache
+from typing import Iterable
+
+import k2
+from k2 import Fsa, SymbolTable
+
+
+class TrainingGraphCompiler:
+    def __init__(self, L: Fsa, vocab: SymbolTable, oov: str = '<UNK>'):
+        self.L = L
```
csukuangfj, line 10: Do we need to add documentation here saying that L is actually L_invert ?

egs/librispeech/asr/simple_v1/decode.py
```diff
@@ -0,0 +1,175 @@
...
+import sys
+import warnings
+from concurrent.futures import ProcessPoolExecutor
+from pathlib import Path
+import math
+
+import editdistance
+import numpy as np
+import torch
+import torchaudio
```
csukuangfj, line 17: `torchaudio` is never used in this file and can be removed.

snowfall/models/interface.py
```diff
@@ -0,0 +1,20 @@
+from torch import nn
+
+
+class AcousticModel(nn.Module):
+    """
+    AcousticModel specifies the common attributes/methods that
```
csukuangfj, line 6: I am afraid these common attributes will not be saved `by save_checkpoint` by default.

danpovey:
Let's aim to merge this quite soon, doesn't have to be perfect.

danpovey:
I get an error with library loading (libc10.so) unless I swap the order of importing k2 and torch, but otherwise it works for me.
The speed is good, at least when compiled in release mode-- less than half of training is spent in k2 stuff, and of that, much is the graph creation which can be cached.
This is from the nsys profile output.  You have to be careful when interpreting it because sometimes an apparently small operation will block for a long time, but it's because it has to wait for previously scheduled operations to complete.
Unfortunately, for a reason I have not yet been able to determine, only the torch kernels show up on the nsys profiling output, and not our kernels, but our nvtx ranges show up.

danpovey:
Incidentally, to get good results we will probably either have to use a different topology where there is a phone-specific blank (could be tied among different word-position-dependent versions, perhaps), or use an LSTM at or near the last layer so we can get a peaky output.

snowfall/training/graph.py
pzelasko: I'll apply the suggestion and keep it - just in case the default ever changes.

snowfall/training/graph.py
pzelasko: The discussion moved here: https://github.com/k2-fsa/k2/pull/476

It seems that since the intersection has to be done (?) on the CPU anyway, it's better to construct each graph on the CPU, and I'd prefer to do it individually for now (otherwise you can get a batch with a mix of cached and non-cached texts and the logic to handle that properly is a bit more complex). I can modify this later if it's still too slow.

The `lru_cache` caches the whole function conditioned on the input - if the argument (text) is present in the cache, it won't execute the function basically.

egs/librispeech/asr/simple_v1/decode.py
pzelasko: I'll wait for @qindazhu to merge his decoding PR and then I'll merge mine to avoid merge conflicts - I can optimize imports again before that.

snowfall/models/interface.py
pzelasko: I added handling them to save/load checkpoint; I think we should make these a part of the model class actually, I can do it in a future PR.

pzelasko:
If you want to merge this quickly, I addressed all the comments and I think it's okay to go (it does work). I'll keep pushing more changes with in the next PRs. I suggest merging #34 first to avoid any merge conflicts.

pzelasko:
Update: I actually went through Haowen's decoding code and made all the changes suggested in the other PR. We can merge this PR and close the other one.

The decoding code runs up to the point of GPU OOM in the intersection, the same as Dan reported in #34.

snowfall/training/graph.py
```diff
@@ -6,21 +6,39 @@
...
-        self.L = L
+    """
+    Compiles and caches the supervision FSAs created for text phrases.
+    The Fsas are created and stored on the CPU and the user is expected to transfer
+    them to GPU themselves (that might change at some point).
+
+    Args:
+        L_inv:
+            An ``Fsa`` that represents the inverted lexicon (L), i.e. has phones as ``symbols``
+            and words as ``aux_symbols``.
```
freewym, line 17: I think it should be reversed, i.e. has word as "symbols" and phones as "aux_symbols"

snowfall/decoding/graph.py
```diff
@@ -0,0 +1,46 @@
...
+        aux_labels_disambig_id_start: int
+) -> Fsa:
+    # if not os.path.exists(lang_dir + '/LG.pt'):
+    #     print("Loading L_disambig.fst.txt")
+    #     with open(lang_dir + '/L_disambig.fst.txt') as f:
+    #         L = k2.Fsa.from_openfst(f.read(), acceptor=False)
+    #     print("Loading G.fsa.txt")
+    #     with open(lang_dir + '/G.fsa.txt') as f:
+    #         G = k2.Fsa.from_openfst(f.read(), acceptor=True)
+    L = k2.arc_sort(L.invert_())
```
freewym, line 20: maybe emphasize it's actually Linv as well

qindazhu:
@pzelasko you can just merge this PR if you think it's ready and I'll close mine. Anyway, the decode code now is just a very early draft, we definitely to debug into it later.

snowfall/training/graph.py
qindazhu: Yeah, agree with @pzelasko here, considering we may shuffle between epoch, cache by text instead of `texts` woulb be much simpler.

danpovey:
Great work!!

PR #40, opened 2020-12-14
Decode fixes

danpovey:


danpovey:
Eventually hit a bug, but producing output for the 1st time.
```
['AND', 'THE', 'AND']
['AND', 'THEN']
['AND', 'THAT', 'THE', 'GROUND', 'AND', 'THE', 'WAS']
[]
['AND', 'THE']
['AND', 'THE', 'MEN', 'OF', 'WHICH', 'HAVE']
['BUT', 'THE', 'PASSION', 'OF', 'THE']
['AND', 'SEE', 'WHAT']
[]
['IT', 'WAS', 'ONE', 'OF', 'THE']
['SECOND', 'WHEN', 'THE']
[]
['AND', 'AND']
['AND', 'THEN']
['BUT', 'WITH']
['THAT', 'WAS', 'WITH']
['SINCE', 'ONE', 'OF', 'THE']
[]
[]
Traceback (most recent call last):
  File "decode.py", line 181, in <module>
    main()
  File "decode.py", line 155, in main
    results = decode(dataloader=test_dl,
  File "decode.py", line 93, in decode
    batch_idx, len(dataloader),
  File "/ceph-dan/.local/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 388, in __len__
    length = self._IterableDataset_len_called = len(self.dataset)  # type: ignore
TypeError: object of type 'K2SpeechRecognitionIterableDataset' has no len()
de-74279-k2-dev-0929153814-7686b86bc5-qbfjq:simple_v1: 

```

pzelasko:
IterableDatasets don't have len() specified due to their "dynamic" nature; you could count the number of cuts processed so far (variable in each batch) and report it with regard to `len(dataset.cuts)`

PR #41, opened 2020-12-15
fix bugs

qindazhu:


qindazhu:
Finally get WER, though it's very bad, not sure if it's because we use too small beam 

`%WER 97.46% [51238 / 52576, 90 ins, 46270 del, 4878 sub ]`

danpovey:
Great!
Blank bias may not be large enough.

qindazhu:
I tried `[-5,0]`, the best result I can get is with `blank_bias = -3.0`
 `%WER 96.73% [50859 / 52576, 492 ins, 41303 del, 9064 sub ]`

PR #45, opened 2020-12-17
[WIP] Position independent phones + LSTM model

pzelasko:


pzelasko:
## Some updates

- I changed position-dependent to position-independent phones and updated the code to automatically choose the output layer size based on the phone symbol table;
- I changed the data augmentation scheme from "WavAugment" default to Kaldi's speed perturbation at 0.9x and 1.1x. My motivation was that I thought maybe it's too difficult for the model to learn with potentially highly reverberated audio. I realized there's an issue in Lhotse while doing that https://github.com/lhotse-speech/lhotse/issues/166
- I added a simple model with 2 conv and 5 lstm layers that has roughly the same number of parameters as the previous one (~12M). There's a lot of room for improvement so feel free to suggest anything.

## Observations

The models have widely different objf values, but very similar (bad) WER.
- TDNN is going down to ~0.77 for train and oscillates between 0.44-0.49 for dev; the WER is 92.1%
- LSTM (without skip connections and batch norm between LSTM layers) goes down to 1.11 for train and 1.33 for dev (much larger!); the WER is 89.3%
- LSTM (with skip connections and batch norm between LSTM layers) goes down to ~0.77 for train and 0.38 for dev (should be better than TDNN); WER is 90.1%

A lot of recognitions are empty (31776 deletions, 52576 reference words in total). I don't think it's related to the utterance length (some longer ones have recognitions).

## Next steps

- I should fix the data aug problem in Lhotse (for now I'll use speed perturbation of 1.05 and 1.1 instead, and re-run).
- I guess it's possible there's still some issue in decoding - maybe it's worth checking a short utterance that had empty hypothesis and visualizing the posteriors and possibly the lattice?
- Any suggestions?

qindazhu:
@pzelasko , I added rm-epsilon for the decoding graph in https://github.com/k2-fsa/snowfall/pull/46  , seems get a little gain on the old TDNN model ( `%WER 96.73%` -> ` 93.81% `), maybe you can try this with your latest scripts (though the gain is very small)

I think after @danpovey get the latest intersect_pruned code ready, maybe we can try with large beam.

pzelasko:
@qindazhu 

with LSTM, speed perturb 1.05 and 1.1 *before* your changes, I got:
```
%WER 89.52% [47068 / 52576, 522 ins, 32163 del, 14383 sub ]
```

after rm-epsilon, it's:
```
%WER 90.97% [47830 / 52576, 829 ins, 30220 del, 16781 sub ]
```

The WER went up but it probably doesn't mean anything. There's less deletions though.

qindazhu:
Ah, OK, thanks.. let's keep the changes though as it does not hurt

PR #46, opened 2020-12-18
remove epsilon of decoding graph

qindazhu:


pzelasko:
LGTM, I'll run it with the models I have

qindazhu:
Merging, also merged the python wrapped code https://github.com/k2-fsa/k2/pull/523, you can just pull the latest k2 master to use this function.

PR #47, opened 2020-12-18
reorder texts by indices

Curisan:
It seems there is a small bug in train.py and decode.py. In train.py：

 supervision_segments = supervision_segments[indices]

    texts = supervisions['text']
    assert feature.ndim == 3
    # print(supervision_segments[:, 1] + supervision_segments[:, 2])
The "texts" is not reorder by indices, while supervision_segments have been reorder. And in decode.py there is same bug.
I reorder "texts" by indices and retrain the model, the WER reduce to 87.01% after 5 epoches.

### DIFF
--- a/egs/librispeech/asr/simple_v1/decode.py
+++ b/egs/librispeech/asr/simple_v1/decode.py
@@ -40,6 +40,7 @@ def decode(dataloader: torch.utils.data.DataLoader, model: AcousticModel,
         indices = torch.argsort(supervision_segments[:, 2], descending=True)
         supervision_segments = supervision_segments[indices]
         texts = supervisions['text']
+        texts = [texts[idx] for idx in indices]
         assert feature.ndim == 3
 
         feature = feature.to(device)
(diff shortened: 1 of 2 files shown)


danpovey:
Fantastic!
I wonder if it might be easier if we reorder our hyps by the inverse of the indices, so that we can process things in the same order as they were given to us?
maybe:
```
    def invert_permutation(indices):
         ans = torch.zeros(indices.shape, device=indices.device, dtype=torch.long)
         ans[indices] = torch.arange(0, indices.shape[0], device=indices.device)
         return ans
```
(note: would only work if indices has long type)

danpovey:
Let me merge for now though, so we have more meaningful WERs.

danpovey:
Now that we fixed this problem we may want to re-check tuning the bias scale.

danpovey:
@csukuangfj is there online documentation for the Python API of class _k2.RaggedInt ?  Or of RaggedShape?

csukuangfj:
> @csukuangfj is there online documentation for the Python API of class _k2.RaggedInt ? Or of RaggedShape?

Doc for `RaggedShape` is added in https://github.com/k2-fsa/k2/pull/532

Will add doc for ` _k2.RaggedInt` later.

PR #53, opened 2020-12-18
Beam changes

danpovey:


danpovey:
With this and the latest k2 it should be possible to get 16% WER.

PR #54, opened 2020-12-18
Sync with Lhotse's speed perturbation fix

pzelasko:


pzelasko:
It runs alright for me, and the model is converging as usual with the first couple hundred of batches. I'm merging - please update your Lhotse to the latest master branch before you attempt to run.

danpovey:
Great!  Do we have to rerun prepare.py in snowfall to get the benefit of
this, and if we don't, will it still run?

pzelasko:
You have to re-run to get the benefit; it will still run if you don't.

pzelasko:
(disambiguation: the training will still run with old prepare.py output; new prepare.py won't run with older Lhotse)

PR #55, opened 2020-12-20
Improve how we extract labels from graphs

danpovey:


danpovey:
Note: this will crash if you don't use the latest version of k2.

danpovey:
... also note: it's taking us around 1 minute to decode.   Assuming there are 5 hours of test dataset (seems to be), that would be about 300xRT !

csukuangfj:
👍

PR #52, opened 2020-12-18
Misc. script change...

danpovey:
s

PR #57, opened 2020-12-21
Support repeated symbols in the output for CTC training.

csukuangfj:
The current supervision graph is generated in the following way:
https://github.com/k2-fsa/snowfall/blob/1a6412259402810f29ef00937b54b6171257b054/snowfall/training/graph.py#L41-L43

It supports repeated blanks in the output but does not support repeated non-blank symbols.

This pull-request tries to support repeated symbols in the output.
For example, the supervision graph for `HELLO  HH AH0 L OW1` is

hello

----

The average `objf` is around `0.09`, lower than that of the master branch, which is about `0.2`
```
2020-12-21 14:05:11,744 INFO [train.py:173] batch 1210, epoch 9/10 global average objf: 0.088675 over 35966885.0 frames (100.0% kept),
current batch average objf: 0.083051 over 29838 frames (100.0% kept) avg time waiting for batch 0.007s
2020-12-21 14:05:18,166 INFO [train.py:173] batch 1220, epoch 9/10 global average objf: 0.088579 over 36257353.0 frames (100.0% kept),
current batch average objf: 0.081375 over 29573 frames (100.0% kept) avg time waiting for batch 0.007s
2020-12-21 14:05:22,260 INFO [common.py:96] Save checkpoint to exp-lstm-adam/epoch-9.pt: epoch=9, learning_rate=0.001, objf=3225266.8044656334
2020-12-21 14:05:22,409 INFO [common.py:133] write training info to exp-lstm-adam/epoch-9-info
2020-12-21 14:05:22,409 WARNING [train.py:347] Done
```

---

The decoding result for epoch 1 with the current decoding graph is 19.41%, which is worse than 16% reported in #52
```
ref= ['ALL', 'IS', 'SAID', 'WITHOUT', 'A', 'WORD']
hyp= ['ALWAYS', 'SAID', 'WITHOUT', 'A', 'WORD']
ref= ['HE', 'COULD', 'WAIT', 'NO', 'LONGER']
hyp= ['HE', 'COULD', 'WAIT', 'NO', 'LONGER']
%WER 19.41% [10203 / 52576, 1648 ins, 913 del, 7642 sub ]
```

The more epochs it runs, the worse the final WER, but the lower objf it gets. The best WER is from epoch 1.

@danpovey  Do you think the bad WER is caused by reusing the existing graph? If so, how do we change the existing
decoding graph?


snowfall/common.py
```diff
@@ -125,3 +127,115 @@ def save_training_info(
...
+    prev_state = 0
+    whitespace = re.compile('[ \t]+')
+    words = whitespace.split(text.strip(' \t\r\n'))
+    rules = ''
+    for word in words:
+        if word not in lexicon:
+            word = oov
+
+        phones = lexicon[word]
+        phones = phones[0]  # TODO(fangjun): handle multiple pronunciations
```
qindazhu, line 223: Maybe worse WER is because you force each word has only one pronunciation here for now?

csukuangfj: thanks, I will try it.

danpovey: You didn't also change the decoding graph to support the same repeated symbols?

danpovey:
We discussed offline.. I think it would be more in the spirit of FSAs to, instead of constructing this manually, to compose with a special FST that allows for the repeated symbols and then the epsilons.

If the non-epsilon symbols are numbered 1... N, we have N+2 states in the FST where state N+1 is the final-state (view this as corresponding to symbol -1); and state 0 is the start-state which also handles epsilon; state i handles symbol i.

For 0 <= i <= N, the i'th state has a self-loop with ilabel:olabel epsilon:i, and for each  -1 <= j <= N it has an arc with ilabel:olabel j:j, going to state j (or to state N+1 if j == -1).

We could in principle create special intersection code for fully-connected cyclic FSAs like this, where the scores and the olabels are stored as Array2's, but for now there's no hurry (at least unless it becomes a limiting factor in graph creation).

snowfall/common.py
csukuangfj: No, I reused the decoding graph from the existing code. Will change the training graph and decoding graph
in the way as you suggested.

Thanks!

egs/librispeech/asr/simple_v1/train.py
```diff
@@ -26,6 +26,7 @@
 from snowfall.models import AcousticModel
 from snowfall.models.tdnn_lstm import TdnnLstm1b
 from snowfall.training.graph import TrainingGraphCompiler
+from snowfall.training.ctc_graph import CtcTrainingGraphCompiler
```
pzelasko, line 29: I think you forgot to check this file in

egs/librispeech/asr/simple_v1/train.py
```diff
@@ -219,29 +220,18 @@ def main():
...
-            L_inv = k2.arc_sort(L.invert_())
-            torch.save(L_inv.as_dict(), lang_dir / 'Linv.pt')
-
-    phone_symbol_table = k2.SymbolTable.from_file(lang_dir / 'phones.txt')
-
-    graph_compiler = TrainingGraphCompiler(
-        L_inv=L_inv,
-        vocab=symbol_table,
-    )
+    graph_compiler = CtcTrainingGraphCompiler(lexicon_txt=str(lexicon_txt),
```
pzelasko, line 227: how about just `lexicon`, `phones` and `words` (or `vocab`/`vocabulary`)? "txt" is just a disk storage detail, not relevant to the concept.

pzelasko: Pass the `Path` object in and use that to read it inside, e.g. there's a nice method `path.read_text()`.

snowfall/common.py
```diff
@@ -75,22 +81,20 @@ def load_checkpoint(filename: Pathlike, model: AcousticModel) -> Tuple[int, floa
...
 
-def save_checkpoint(
-        filename: Pathlike,
-        model: AcousticModel,
-        epoch: int,
-        learning_rate: float,
-        objf: float,
-        local_rank: int = 0
-) -> None:
+def save_checkpoint(filename: Pathlike,
```
pzelasko, line 84: There is a lot of formatting changes that obfuscate what you actually did. I guess we need common auto-formatting script or settings. Preferably something that can work easily with IDEs like PyCharm (not sure what you use).

snowfall/common.py
```diff
@@ -125,3 +127,115 @@ def save_training_info(
         f.write('best epoch: {}\n'.format(best_epoch))
 
     logging.info('write training info to {}'.format(filename))
+
+
+def read_lexicon(filename: str) -> Dict[str, List[List[str]]]:
```
pzelasko, line 132: best to use `Path` for paths (or `Pathlike` if the user is likely to pass `str` in anyway)

snowfall/common.py
```diff
@@ -125,3 +127,115 @@ def save_training_info(
...
+      lexicon:
+        It is returned by :func:`read_lexicon`.
+      text:
+        The transcript. It contains words separated by spaces.
+      phone_symbol_table:
+        Phone symbol table.
+      word_symbol_table:
+        Word symbol table.
+    '''
+    oov = '<UNK>'
```
pzelasko, line 211: Should be a parameter

snowfall/common.py
csukuangfj: Sorry for the inconvenience.

I am using vim configured with `yapf==0.27.0` for auto-formatting. The configure file for yapf is from
https://github.com/k2-fsa/k2/blob/master/.style.yapf

What is the formatter that you are using in PyCharm?

snowfall/common.py
csukuangfj: thanks, will change.

snowfall/common.py
pzelasko: I'm using the built-in formatter, but PyCharm can be configured to use something external (e.g. I used https://github.com/psf/black in the past for one project).

csukuangfj: I can switch to `black` in vim. Is there any config file for black? Which version are you using?

pzelasko: For config check this out: https://github.com/psf/black#configuration-format 

I am not using any particular version at the moment, I guess any one is fine (we should freeze it in the requirements).

snowfall/common.py
csukuangfj: Resolved by removal.

snowfall/common.py
csukuangfj: Resolved by removal.

egs/librispeech/asr/simple_v1/train.py
csukuangfj: Resolved by removal.

egs/librispeech/asr/simple_v1/train.py
csukuangfj: Thanks, added.

csukuangfj:
WER for epoch 7 (counting from 0) is `14.43%`.

```
ref=['HE', 'TRIED', 'TO', 'THINK', 'HOW', 'IT', 'COULD', 'BE']                                                                [10/1935]
hyp=['HE', 'TRIED', 'TO', 'THINK', 'HOW', 'IT', 'COULD', 'BE']
ref=['STEPHANOS', 'DEDALOS']
hyp=['STEP', 'AND', 'ALL', 'DAEDALUS']
ref=['THE', 'UNIVERSITY']
hyp=['THE', 'UNIVERSITY']
ref=['ALL', 'IS', 'SAID', 'WITHOUT', 'A', 'WORD']
hyp=['ALWAYS', 'SAID', 'WITHOUT', 'A', 'WORD']
ref=['HE', 'COULD', 'WAIT', 'NO', 'LONGER']
hyp=['HE', 'COULD', 'WAIT', 'NO', 'LONGER']

2020-12-22 19:22:30,319 INFO [decode.py:210] %WER 14.43% [7585 / 52576, 1048 ins, 640 del, 5897 sub ]
```
The decoding graph for inference is not changed. Only the supervision graph for training is changed to handle
repeated non-blank symbols.

---

I think the reason for the bad WER of the previous commit is that:
 - It did not use optional silence in the training stage, but it added an optional silence to the beginning and end of each word
    to the decoding graph during inference
 - It considered only one pronunciation of a word if there are multiple alternate pronunciations.

csukuangfj:
Here is a sample topology for absorbing repeated symbols and its composition result with a lexicon, which is essentially
the same with the one constructed manually.

Screen Shot 2020-12-22 at 7 45 01 PM

Screen Shot 2020-12-22 at 7 45 07 PM

Screen Shot 2020-12-22 at 7 45 16 PM

snowfall/training/ctc_graph.py
```diff
@@ -0,0 +1,159 @@
...
+    '''Compose two graphs.
+
+    Args:
+      a_fsa:
+        The FSA on the left hand side.
+      b_fsa:
+        The FSA on the right hand side.
+
+    Returns:
+      The composition result. It will have:
```
qindazhu, line 51: `It will have:` what? seems  there were some documentation?

snowfall/training/ctc_graph.py
```diff
@@ -0,0 +1,159 @@
...
+                rules += f'{i} {i} {phones[i]} 0 0.0\n'
+            else:
+                rules += f'{i} {j} {phones[j]} {phones[j]} 0.0\n'
+        rules += f'{i} {final_state} -1 -1 0.0\n'
+    rules += f'{final_state}'
+    ans = k2.Fsa.from_str(rules)
+    return ans
+
+
+def compose(a_fsa: k2.Fsa, b_fsa: k2.Fsa) -> k2.Fsa:
```
qindazhu, line 41: I was thinking we would write this function in `k2` instead of here.

snowfall/training/ctc_graph.py
```diff
@@ -0,0 +1,159 @@
...
+    Iterable,
+    List,
+)
+
+import torch
+import k2
+import _k2
+
+
+def build_ctc_topo(phones: List[int]) -> k2.Fsa:
```
pzelasko, line 14: Would be more generic as: phones -> tokens

snowfall/training/ctc_graph.py
pzelasko: +1

snowfall/training/ctc_graph.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self,
+                 L_inv: k2.Fsa,
+                 phones: k2.SymbolTable,
+                 words: k2.SymbolTable,
+                 oov: str = '<UNK>'):
+        '''
+        Args:
+          L_inv:
+            Its labels are words, while the aux_labels are phones. Must
+            be arc sorted.
```
pzelasko, line 125: Would be better to arc sort in __init__, more user friendly and shouldn’t impact the performance

snowfall/training/ctc_graph.py
```diff
@@ -0,0 +1,159 @@
...
+            setattr(out_fsa, name, a_value)
+
+    for name, b_value in b_fsa.named_non_tensor_attr():
+        if not hasattr(out_fsa, name):
+            setattr(out_fsa, name, b_value)
+
+    return out_fsa
+
+
+class CtcTrainingGraphCompiler:
```
pzelasko, line 114: BTW I think you can just remove the previous graph compiler code as it was essentially a half-baked CTC, I don’t think we’ll need it going forward?

egs/librispeech/asr/simple_v1/train.py
```diff
@@ -25,7 +25,7 @@
 from snowfall.common import setup_logger
 from snowfall.models import AcousticModel
 from snowfall.models.tdnn_lstm import TdnnLstm1b
-from snowfall.training.graph import TrainingGraphCompiler
+from snowfall.training.ctc_graph import CtcTrainingGraphCompiler as TrainingGraphCompiler
```
pzelasko, line 28: I think that “as” obfuscates the intent in code below (IMO it’s a good idea to have the specific topology type in the class name where it’s used)

snowfall/training/ctc_graph.py
```diff
@@ -0,0 +1,159 @@
...
+
+from functools import lru_cache
+from typing import (
+    Iterable,
+    List,
+)
+
+import torch
+import k2
+import _k2
```
pzelasko, line 11: I’m not sure but importing _k2 outside of k2 code base seems like an antipattern, maybe something needs to be included in public k2 API?

snowfall/training/ctc_graph.py
csukuangfj: Fixed.

snowfall/training/ctc_graph.py
csukuangfj: I wrote [one][1] some time ago in k2 but it was later removed.

https://github.com/k2-fsa/k2/pull/282#discussion_r513161115

https://github.com/k2-fsa/k2/pull/282#discussion_r513517896

[1]: https://github.com/k2-fsa/k2/commit/8538bd1e9a80e4a2835b4c189212bfc7a3e6e392

snowfall/training/ctc_graph.py
csukuangfj: Renamed to `tokens`.

danpovey:
Great work!!  Since I'm  away for a few days, you guys merge if you feel it's best...
Would be nice to see if modifying the decoding graph the same way is helpful, but not a big hurry on that,  you can merge as-is since it seems helpful.
It's nice that we're able to do the whole pipeline in the spirit of FSAs without too much manual work.

snowfall/training/ctc_graph.py
csukuangfj: I agree.

If we can move `compose` to k2, then we do not need `import _k2`.

snowfall/training/ctc_graph.py
danpovey: I guess this can be moved to k2 for now.

csukuangfj:
> Would be nice to see if modifying the decoding graph the same way is helpful,

I tried it but `k2.determinize` on `compose(ctc_topo, LG)`  never terminated.

danpovey:
compose after determinizing

csukuangfj:
> compose after determinizing


After determinizing, the `aux_labels` is a ragged tensor, which is not supported by `compose`.
`compose` expects an`aux_labels` with  type `torch.Tensor`

csukuangfj:
> compose after determinizing


Will try it.

csukuangfj:
Tried to replace
https://github.com/k2-fsa/snowfall/blob/1a6412259402810f29ef00937b54b6171257b054/snowfall/decoding/graph.py#L58
with

```python
LG = k2.compose(ctc_topo, LG)
LG = k2.connect(LG)
```

`compose` ran succesfully but there was an error from `k2::MultiGraphDenseIntersectPruned::MultiGraphDenseIntersectPruned`:

```
About to decode
[F] /.../k2/k2/csrc/intersect_pruned.cu:k2::MultiGraphDenseIntersectPruned::MultiGraphDenseIntersectPruned(k2::Fsa
Vec&, k2::DenseFsaVec&, float, float, int32_t, int32_t):171 Check failed: num_keys == (uint32_t)num_keys
```

Seems that `num_keys` is overflowed.

danpovey:
reduce num_frames in.decode

csukuangfj:
> reduce num_frames in.decode


Thanks, seems working. Got a WER of `12.96%`.

csukuangfj:
This pull-request requires https://github.com/k2-fsa/k2/pull/543

Will run the pipeline from scratch to see whether the result can be reproduced.

danpovey:
thanks.
when i get back ill extend that code to handle larger.sizes.

egs/librispeech/asr/simple_v1/train.py
csukuangfj: Fixed as you suggested.

snowfall/training/ctc_graph.py
csukuangfj: Removed.

snowfall/common.py
csukuangfj: Fixed.

snowfall/training/ctc_graph.py
csukuangfj: Removed to k2.

snowfall/training/ctc_graph.py
csukuangfj: Fixed.

snowfall/training/ctc_graph.py
csukuangfj: Fixed.

csukuangfj:
I confirm that the result can be reproduced from a fresh start (trained for 10 epochs):
```
ref=['AND', 'LOVE', 'BE', 'FALSE']
hyp=['AND', 'THE', 'LOVE', 'BE', 'FALSE']
ref=['ALL', 'IS', 'SAID', 'WITHOUT', 'A', 'WORD']
hyp=['ALWAYS', 'SAID', 'WITHOUT', 'A', 'WORD']
ref=['I', 'LOVE', 'THEE', 'WITH', 'A', 'LOVE', 'I', 'SEEMED', 'TO', 'LOSE', 'WITH', 'MY', 'LOST', 'SAINTS', 'I', 'LOVE', 'THEE', 'WITH', 'THE', 'BREATH', 'SMILES', 'TEARS', 'OF', 'ALL', 'MY', 'LIFE', 'AND', 'IF', 'GOD', 'CHOOSE', 'I', 'SHALL', 'BUT', 'LOVE', 'THEE', 'BETTER', 'AFTER', 'DEATH']
hyp=['I', 'LOVE', 'THEE', 'WITH', 'A', 'LOVE', 'I', 'SEEMED', 'TO', 'LOSE', 'WITH', 'MY', 'LAST', 'SAINTS', 'I', 'LOVE', 'THEE', 'WITH', 'THE', 'BREATH', 'SMILES', 'HEARS', 'OF', 'ALL', 'MY', 'LIFE', 'AND', 'IF', 'GOD', 'CHOOSE', 'I', 'SHALL', 'BUT', 'LOVE', 'THE', 'BETTER', 'AFTER', 'DEATH']

2020-12-23 20:09:55,264 INFO [decode.py:214] %WER 12.97% [6817 / 52576, 897 ins, 654 del, 5266 sub ]
```

---

All comments are resolved. Merging.

danpovey:
thanks!  great work!

fanlu:
I have tested this pr on [aishell](https://www.openslr.org/33/) dataset, and get the result below
```
INFO [decode.py:214] %WER 23.62% [15221 / 64428, 1372 ins, 1847 del, 12002 sub ]
```

danpovey:
Great!
Is there any chance you could make a pull request?

PR #70, opened 2021-01-06
rename asg to mmi_bigram

csukuangfj:


danpovey:
Looks OK to me.  
Motivation is that while, like ASG, we're jointly training the probs with the model, we actually use a different objective, chosen to keep the LM probabilistic.  So not sure that it makes sense to call it ASG, plus MMI has a much longer history.

csukuangfj:
Merging

PR #78, opened 2021-01-20
Make ctc topo normal.

yaguanghu:
#77 
@danpovey @csukuangfj 
Please have a look at it.

### DIFF
--- a/egs/aishell/asr/simple_v1/decode.py
+++ b/egs/aishell/asr/simple_v1/decode.py
@@ -137,8 +137,9 @@ def main():
     lang_dir = Path('data/lang_nosp')
     symbol_table = k2.SymbolTable.from_file(lang_dir / 'words.txt')
     phone_symbol_table = k2.SymbolTable.from_file(lang_dir / 'phones.txt')
-    ctc_topo = k2.arc_sort(build_ctc_topo(list(
-        phone_symbol_table._id2sym.keys())))
+    phone_ids = get_phone_symbols(phone_symbol_table)
+    phone_ids_with_blank = [0] + phone_ids
+    ctc_topo = k2.arc_sort(build_ctc_topo(phone_ids_with_blank))
 
     if not os.path.exists(lang_dir / 'LG.pt'):
         print("Loading L_disambig.fst.txt")
--- a/egs/aishell/asr/simple_v1/decode.py
+++ b/egs/aishell/asr/simple_v1/decode.py
@@ -137,8 +137,8 @@ def main():
     lang_dir = Path('data/lang_nosp')
     symbol_table = k2.SymbolTable.from_file(lang_dir / 'words.txt')
     phone_symbol_table = k2.SymbolTable.from_file(lang_dir / 'phones.txt')
-    ctc_topo = build_ctc_topo(list(phone_symbol_table._id2sym.keys()))
-    ctc_topo = k2.arc_sort(ctc_topo)
+    ctc_topo = k2.arc_sort(build_ctc_topo(list(
+        phone_symbol_table._id2sym.keys())))
 
     if not os.path.exists(lang_dir / 'LG.pt'):
         print("Loading L_disambig.fst.txt")
--- a/snowfall/decoding/graph.py
+++ b/snowfall/decoding/graph.py
@@ -19,7 +19,9 @@ def compile_LG(L: Fsa, G: Fsa, ctc_topo: Fsa, labels_disambig_id_start: int,
         G:
             An ``Fsa`` that represents the language model (G), i.e. it's an acceptor
             with words as ``symbols``.
-        ctc_topo:  Epsilons are in `labels` and `aux_labels` contain phone IDs.
+        ctc_topo:  CTC topology fst, When 0 appears on the left side, it represents
+                   the blank symbol; when it appears on the right side,
+                   it indicates an epsilon.
         labels_disambig_id_start:
             An integer ID corresponding to the first disambiguation symbol in the
             phonetic alphabet.
--- a/snowfall/decoding/graph.py
+++ b/snowfall/decoding/graph.py
@@ -19,7 +19,7 @@ def compile_LG(L: Fsa, G: Fsa, ctc_topo: Fsa, labels_disambig_id_start: int,
         G:
             An ``Fsa`` that represents the language model (G), i.e. it's an acceptor
             with words as ``symbols``.
-        ctc_topo:  CTC topology fst, When 0 appears on the left side, it represents
+        ctc_topo:  CTC topology fst, in which when 0 appears on the left side, it represents
                    the blank symbol; when it appears on the right side,
                    it indicates an epsilon.
         labels_disambig_id_start:
--- a/snowfall/decoding/graph.py
+++ b/snowfall/decoding/graph.py
@@ -5,7 +5,7 @@ import torch
 from k2 import Fsa
 
 
-def compile_LG(L: Fsa, G: Fsa, ctc_topo_inv:Fsa, labels_disambig_id_start: int,
+def compile_LG(L: Fsa, G: Fsa, ctc_topo: Fsa, labels_disambig_id_start: int,
                aux_labels_disambig_id_start: int) -> Fsa:
     """
     Creates a decoding graph using a lexicon fst ``L`` and language model fsa ``G``.
@@ -19,7 +19,7 @@ def compile_LG(L: Fsa, G: Fsa, ctc_topo_inv:Fsa, labels_disambig_id_start: int,
         G:
             An ``Fsa`` that represents the language model (G), i.e. it's an acceptor
             with words as ``symbols``.
-        ctc_topo_inv:  Epsilons are in `aux_labels` and `labels` contain phone IDs.
+        ctc_topo:  Epsilons are in `labels` and `aux_labels` contain phone IDs.
         labels_disambig_id_start:
             An integer ID corresponding to the first disambiguation symbol in the
             phonetic alphabet.
@@ -47,7 +47,8 @@ def compile_LG(L: Fsa, G: Fsa, ctc_topo_inv:Fsa, labels_disambig_id_start: int,
     if isinstance(LG.aux_labels, torch.Tensor):
         LG.aux_labels[LG.aux_labels >= aux_labels_disambig_id_start] = 0
     else:
-        LG.aux_labels.values()[LG.aux_labels.values() >= aux_labels_disambig_id_start] = 0
+        LG.aux_labels.values()[
+            LG.aux_labels.values() >= aux_labels_disambig_id_start] = 0
     logging.debug("Removing epsilons")
     LG = k2.remove_epsilons_iterative_tropical(LG)
     logging.debug(f'LG shape = {LG.shape}')
@@ -60,7 +61,7 @@ def compile_LG(L: Fsa, G: Fsa, ctc_topo_inv:Fsa, labels_disambig_id_start: int,
     LG = k2.arc_sort(LG)
 
     logging.debug("Composing")
-    LG = k2.compose(ctc_topo_inv, LG)
+    LG = k2.compose(ctc_topo, LG)
 
     logging.debug("Connecting")
     LG = k2.connect(LG)
--- a/snowfall/training/ctc_graph.py
+++ b/snowfall/training/ctc_graph.py
@@ -10,14 +10,12 @@ import k2
 
 def build_ctc_topo(tokens: List[int]) -> k2.Fsa:
     '''Build CTC topology.
-
-    The resulting topology converts repeated input
-    symbols to a single output symbol.
-
-    Caution:
-      The resulting topo is an FST. Epsilons are on the right
-      side (i.e., olabels) and tokens are on the left side (i.e., ilabels)
-
+    A token which appears once on the right side (i.e. olabels) may
+    appear multiple times on the left side (ilabels), possibly with
+    epsilons in between.
+    When 0 appears on the left side, it represents the blank symbol;
+    when it appears on the right side, it indicates an epsilon. That
+    is, 0 has two meanings here.
     Args:
       tokens:
         A list of tokens, e.g., phones, characters, etc.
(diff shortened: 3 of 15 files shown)


snowfall/training/ctc_graph.py
```diff
@@ -15,8 +15,8 @@ def build_ctc_topo(tokens: List[int]) -> k2.Fsa:
     symbols to a single output symbol.
 
     Caution:
-      The resulting topo is an FST. Epsilons are on the left
-      side (i.e., ilabels) and tokens are on the right side (i.e., olabels)
+      The resulting topo is an FST. Epsilons are on the right
```
qindazhu, line 18: It would be better if we can sync the documentation with PR https://github.com/k2-fsa/k2/pull/602/files where @danpovey and I left some comments and @csukuangfj has fixed them.

yaguanghu: Sure, I will modify it when @csukuangfj's PR is ready.

egs/aishell/asr/simple_v1/decode.py
```diff
@@ -137,8 +137,8 @@ def main():
     lang_dir = Path('data/lang_nosp')
     symbol_table = k2.SymbolTable.from_file(lang_dir / 'words.txt')
     phone_symbol_table = k2.SymbolTable.from_file(lang_dir / 'phones.txt')
-    ctc_topo = build_ctc_topo(list(phone_symbol_table._id2sym.keys()))
-    ctc_topo = k2.arc_sort(ctc_topo)
+    ctc_topo = k2.arc_sort(build_ctc_topo(list(
```
csukuangfj, line 140: Please update it like the one in librispeech, which excludes disambiguation symbols from 
`phone_symbol_table`. 

Sorry that aishell has not been kept in sync with librispeech. We have been doing
experiments with the librispeech recipe.

yaguanghu: will do that.

csukuangfj: Like the one in
https://github.com/k2-fsa/snowfall/blob/1567a6553af8519e52fa86a2c6fd5aa6bf0c43f9/egs/librispeech/asr/simple_v1/mmi_bigram_decode.py#L210

snowfall/decoding/graph.py
```diff
@@ -19,7 +19,7 @@ def compile_LG(L: Fsa, G: Fsa, ctc_topo_inv:Fsa, labels_disambig_id_start: int,
         G:
             An ``Fsa`` that represents the language model (G), i.e. it's an acceptor
             with words as ``symbols``.
-        ctc_topo_inv:  Epsilons are in `aux_labels` and `labels` contain phone IDs.
+        ctc_topo:  Epsilons are in `labels` and `aux_labels` contain phone IDs.
```
csukuangfj, line 22: I think `epsilons` are in `aux_labels`.

snowfall/training/ctc_graph.py
```diff
@@ -69,8 +69,7 @@ def __init__(self,
         self.phones = phones
         self.words = words
         self.oov = oov
-        ctc_topo_inv = build_ctc_topo(list(phones._id2sym.keys())).invert_()
-        self.ctc_topo_inv = k2.arc_sort(ctc_topo_inv)
+        self.ctc_topo = k2.arc_sort(build_ctc_topo(list(phones._id2sym.keys())))
```
csukuangfj, line 72: Please also fix this one. Disambiguation symbols should be excluded.

snowfall/decoding/graph.py
qindazhu, line 22: I'm confused by this line, I was thinking the left side, i.e. labels,  would contain `blank` instead of `epsilon`

snowfall/training/mmi_graph.py
```diff
@@ -106,7 +106,7 @@ def __init__(self,
         ctc_topo = build_ctc_topo(phone_symbols_with_blank)
         assert ctc_topo.requires_grad is False
 
-        self.ctc_topo = ctc_topo
+        self.ctc_topo_inv = ctc_topo.invert_()
```
csukuangfj, line 109: Maybe you need to do `k2.arc_sort(ctc_topo.invert_())`.

egs/aishell/asr/simple_v1/decode.py
```diff
@@ -137,8 +137,9 @@ def main():
     lang_dir = Path('data/lang_nosp')
     symbol_table = k2.SymbolTable.from_file(lang_dir / 'words.txt')
     phone_symbol_table = k2.SymbolTable.from_file(lang_dir / 'phones.txt')
-    ctc_topo = build_ctc_topo(list(phone_symbol_table._id2sym.keys()))
-    ctc_topo = k2.arc_sort(ctc_topo)
+    phone_ids = get_phone_symbols(phone_symbol_table)
```
csukuangfj, line 140: Could you also update *train.py* to exclude disambiguation symbols while
building the model?

https://github.com/k2-fsa/snowfall/blob/add948544da04ac4bb9c14cef0057c6969c1b9fb/egs/librispeech/asr/simple_v1/train.py#L298

https://github.com/k2-fsa/snowfall/blob/add948544da04ac4bb9c14cef0057c6969c1b9fb/egs/aishell/asr/simple_v1/train.py#L300

The correct one should look like the following:
https://github.com/k2-fsa/snowfall/blob/add948544da04ac4bb9c14cef0057c6969c1b9fb/egs/librispeech/asr/simple_v1/mmi_bigram_train.py#L334-L336

yaguanghu: sure.

csukuangfj:
Could you verify that the changes do not break the training and decoding?

yaguanghu:
I don't know how to verify it. Do you have any reasonable results for reference?

csukuangfj:
> I don't know how to verify it.

It can be verified by executing `run.sh`. It should give you a WER at the end. For librispeech, the value is between 10.0% and 11.0%.

If you have run it before,
- For the training part, it should converge like your previous runs
- For the decoding part, it should give the same WER as your previous checkpoint

qindazhu:
You can just run `run.sh` (by default it runs`mmi_bigram_*`, so you may need to run `train/decode.py` seperately), the WER of decoding should be in 10%~12%. Noted the WER range is for LibriSpeech, I never run experiments for aishell with latest k2 changes, sorry.

BTW, it would be great if we can rename `train/decode.py` to `ctc_train.py` and `ctc_test.py`

yaguanghu:
OK, I will verify the results and refine the training scripts.

yaguanghu:
I have run the training and decoding pipeline, the final WER is 13.59%. It seems a little worse than your result.
I noticed that the default training script uses only train-clean-100 for training, what's your training data?

csukuangfj:
We have been using run.sh and the same settings in prepare.py.  

Thursday, 21 January 2021, 20:00 +0800 from notifications@github.com  <notifications@github.com>:

danpovey:
It uses musan, right?  And we are talking about the librispeech example?

danpovey:
Possibly there is just a big scatter of results.  If so we should
definitely investigate the reasons.
What I'd like to see is a comparison of the norms of different parameter
matrices of the trained models.
If a particular layer is getting unstable, it may reveal it.
IDK what diagnostics are available from the logs we automatically write?
E.g. convergence plots?

yaguanghu:
I ran the whole pipeline again, and the final WER is 13.23%. I think the first thing we need to do is eliminating randomness, so that the final results is trustable.
BTW I have some problems in running the original pipeline, but I will figure it out and make a comparsion with this change.

danpovey:
Did you run
python3 setup.py install
before running?

yaguanghu:
I ran `pip install -e .`

snowfall/training/ctc_graph.py
```diff
@@ -69,8 +67,9 @@ def __init__(self,
         self.phones = phones
         self.words = words
         self.oov = oov
-        ctc_topo_inv = build_ctc_topo(list(phones._id2sym.keys())).invert_()
-        self.ctc_topo_inv = k2.arc_sort(ctc_topo_inv)
+        phone_ids = get_phone_symbols(phones)
```
danpovey, line 70: A difference here is that ctc_topo no longer includes the disambig symbols.
Of course it's not right to treat the disambig symbols as if they were phones.
But doesn't L contain disambig symbols on its input side?  Perhaps what is happening is that we are eliminating paths with disambig symbols.   We might need to add self-loop with each of the disambig symbols, on all of the states of the ctc topo.

snowfall/training/ctc_graph.py
```diff
@@ -85,9 +84,10 @@ def compile_one_and_cache(self, text: str) -> k2.Fsa:
         tokens = (token if token in self.words else self.oov
                   for token in text.split(' '))
         word_ids = [self.words[token] for token in tokens]
-        fsa = k2.linear_fsa(word_ids)
-        decoding_graph = k2.connect(k2.intersect(fsa, self.L_inv)).invert_()
+        label_graph = k2.linear_fsa(word_ids)
+        decoding_graph = k2.connect(k2.intersect(label_graph,
+                                                 self.L_inv)).invert_()
```
danpovey, line 89: Actually, instead of adding the disambig symbols into the ctc_topo, we can handle them here.
  (i) we can use compose instead of intersection and double inversion, here.
  (ii) after intersecting, we can replace all the disambig symbols with epsilon and then remove epsilons, and then connect.

danpovey: Oh, wait... now I see that train.py is loading L.fst not L_disambig.fst.  So there may be no disambig symbols to deal with.

danpovey:
I suspect there is a bug but don't know what it is.  Perhaps we could start by printing the compiled CTC training graphs for the first minibatch, for this and the unmodified version, and see if there is a difference.
And same with the decoding graph.  (Careful, it is cached by default).

snowfall/training/ctc_graph.py
yaguanghu: Now the L graph contain disambig symbols on its input side. That might be a problem in the current ctc topology. I think it makes sense to add  self-loop with each disambig symbols.

danpovey: I think I was mistaken about this L having disambig symbols, we are loading L.fst in train.py, which lacks them.

yaguanghu: I will check it. If there are no disambig symbols in L.fst, there may be problems while decoding.

danpovey:
We use L_disambig.fst to create the decoding graph.  But the disambig
symbols are removed prior to composing with ctc_topo.
Bear in mind that it should be possible to train a model with one version
of the code and decode with another; that will help narrow
the problem down to training or decoding.

csukuangfj:
I started to re-run the LibriSpeech recipe from scratch using the master branch this morning. Here is the WER

```
2021-01-22 11:49:11,000 INFO [mmi_bigram_decode.py:296] %WER 10.35% [5443 / 52576, 772 ins, 479 del, 4192 sub ]
```

yaguanghu:
> I started to re-run the LibriSpeech recipe from scratch using the master branch this morning. Here is the WER
> 
> ```
> 2021-01-22 11:49:11,000 INFO [mmi_bigram_decode.py:296] %WER 10.35% [5443 / 52576, 772 ins, 479 del, 4192 sub ]
> ```

Aren't we talking about CTC training? My result is based on CTC pipeline.

danpovey:
Ah, that may explain it, does anyone remember what the CTC WER should be?  I would have thought it would be a little better than that, especially now that we have augmentation.

We should test the MMI branch too, because that also uses the ctc topo.  However, I think it would make more sense to modify mmi_graph.py so that it stores ctc_topo instead of ctc_topo_inv, and uses compose instead of intersect then invert.

csukuangfj:
From https://github.com/k2-fsa/snowfall/pull/57#issuecomment-750297209

> 2020-12-23 20:09:55,264 INFO [decode.py:214] %WER 12.97% [6817 / 52576, 897 ins, 654 del, 5266 sub ]

It was trained without MUSAN.

danpovey:
Ah OK.
Can we start putting the WERs at the top of the training scripts?
(mentioning which decode script we used, and when, who ran it,
what changed if there are multiple WERs listed, etc.)?
Yaguang, were you using MUSAN?
I assume yes, if you ran the whole thing from scratch.
Dan

yaguanghu:
It seems like the default ctc training script is now using musan for data augmentation, althoug it prepared musan data.
I will run it again with musan augmentation.

csukuangfj:
WER of CTC training from scratch using the master branch

```
2021-01-22 16:07:01,722 INFO [decode.py:214] %WER 12.76% [6709 / 52576, 901 ins, 747 del, 5061 sub ]
```

csukuangfj:
> It seems like the default ctc training script is now using musan for data augmentation

I have been assuming that your changes are based on the latest master and your code is using MUSAN to verify
the changes.

yaguanghu:
WER of CTC training with musan using this change.
`INFO [ctc_decode.py:216] %WER 12.47% [6555 / 52576, 802 ins, 708 del, 5045 sub ]`

danpovey:
great!we.can merge?  fangjun pls merge if ok w.you.

yaguanghu:
Add the result of master branch ran by me, note this result is trained without musan, I'm also training CTC with musan using master branch.
`INFO [decode.py:214] %WER 13.21% [6946 / 52576, 909 ins, 705 del, 5332 sub ]`

csukuangfj:
+2

Friday, 22 January 2021, 17:26 +0800 from notifications@github.com  <notifications@github.com>:

xiaohui-zhang:
> Add the result of master branch ran by me, note this result is trained without musan, I'm also training CTC with musan using master branch.
> `INFO [decode.py:214] %WER 13.21% [6946 / 52576, 909 ins, 705 del, 5332 sub ]`

Hey Yaguang, can you add this CTC result (or the one above, whichever should be re-produced by librispeech/run.sh w/ CTC's train.py) to the RESULTS file, before the most recent MMI+MBR numbers? Maybe add MMI numbers also if you it's easy for you to get. Keeping track of these will make debug easier in future.

PR #79, opened 2021-01-22
change with k2,use lhotse to prepare and reduce recipe code

pingfengluo:
aishell recipe : change with latest k2,use lhotse to prepare and reduce code


egs/aishell/asr/simple_v1/prepare.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright (c)  2020  Xiaomi Corporation (authors: Junbo Zhang, Haowen Qiu)
+# Copyright (c)  2020  Xiaomi Corporation (authors: Junbo Zhang, Haowen Qiu, Pingfeng Luo)
```
csukuangfj, line 3: Are you also from Xiaomi? Both of the authors are from Xiaomi. That's why their
names are listed together.

danpovey:
Looks good to me, just clarify the copyright.  Could list your name by itself, or if this is work for a company on company time, list something like:
      Copyright  2021   Company Name (author: Pingfeng Luo)
although depending on the corporate culture (i.e. how lawyer-ish they are on copyright issues) you might want to check first.  normally would  be OK.

pingfengluo:
ok.I will modify it.

PR #84, opened 2021-01-28
MMI + MBR training

csukuangfj:
~working in progress.~

---

Ready for review.

PR #86, opened 2021-02-03
[Not for merge] MBR test.

csukuangfj:
See https://github.com/k2-fsa/snowfall/pull/84#issuecomment-772185201

> OK. Well at least changing that weight has more effect on the MMI loss than on the MBR loss, but those loss values don't give me great confidence that the autograd is correct for the MBR loss. Can you try an experiment where in the 1st epoch it's mmi_loss + mbr_loss, and afterward it's just mbr_loss? (You can still plot the mmi loss). I want to see whether it just goes crazy.
I want to develop some automated testing for the autograd, have to think about it..

What this pull-request does is:
- for epoch == 0, total_loss = mmi_loss + mbr_loss
- for epoch >= 1, total_loss = mbr_loss

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_mbr_train.py
+++ b/egs/librispeech/asr/simple_v1/mmi_mbr_train.py
@@ -36,6 +36,7 @@ from snowfall.training.mmi_graph import create_bigram_phone_lm
 from snowfall.training.mmi_mbr_graph import MmiMbrTrainingGraphCompiler
 
 den_scale = 1.0
+g_counter = 0 # for debugging only
 
 
 def get_tot_objf_and_num_frames(tot_scores: torch.Tensor,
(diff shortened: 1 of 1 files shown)


csukuangfj:
## Loss curves

https://tensorboard.dev/experiment/jvlfa7oyRQiDxvO3hPTXZg/#scalars&_smoothingWeight=0

After epoch 0, `mmi_loss` increases as it does not contribute to the total loss. `mbr_loss` does not change too much after epoch 0.

Screen Shot 2021-02-03 at 21 01 27
Screen Shot 2021-02-03 at 21 01 39

danpovey:
OK, looks reasonable.

csukuangfj:
@danpovey 
I have some doubts about `mbr_loss`. It only starts to decrease after epoch 7, when the learning rate starts to decrease.
From epoch 1 to epoch 6, `mbr_loss` is nearly kept the same.

Since the code tries to minimize `mbr_loss`, is there a reason why `mbr_loss` does not decrease
during epoch 1 to epoch 6?

danpovey:
The gradients for the MBR part of the loss can be a little spiky, and I think it's not as easy to optimize as the MMI part.

danpovey:
Could be gradient noise throwing it off.
We should create some randomized gradient checks for the whole procedure though.

PR #92, opened 2021-02-05
Finish code of normalization idea..

danpovey:


danpovey:
this is not for merge, an idea that didn't help, wanted to show someone...

danpovey:
Note on results with this code: WER was worse, e.g. went from 10.2 to 11.4, but the train-valid difference in objective function may have been reduced, which is kind of as expected.
E.g. by epoch 9 with this code:
```
2021-02-05 09:06:51,198 INFO [common.py:91] Save checkpoint to exp-lstm-adam-mmi-bigram-musan/epoch-9.pt: epoch=9, learning_rate=0.0005120000000000001, objf=0.14944308555614835, valid_objf=0.16140129996080735
```
whereas with a different version of the code where the sign of the extra part of the derivative was accidentally reversed, the train-valid gap was greater:
```
2021-02-04 19:12:39,641 INFO [common.py:91] Save checkpoint to exp-lstm-adam-mmi-bigram-musan/epoch-9.pt: epoch=9, learning_rate=0.0005120000000000001, objf=0.12242732541252148, valid_objf=0.16058101499761657
```

snowfall/models/normalize.py
```diff
@@ -0,0 +1,75 @@
+from typing import Optional
+
+import random
+import torch
+from torch import Tensor
+from torch import nn
+
+
+class GradientNormalizeIn(nn.Module):
```
pzelasko, line 9: I think this could use a train/test mode so that you don't add the noise during inference (might be also a reason for WER getting worse)

snowfall/models/normalize.py
```diff
@@ -0,0 +1,75 @@
...
+        x1 = x[0:b]
+        x2 = x[b:2*b]
+
+        # sumsq_diff will be for each sequence.
+        sumsq_diff = ((x1 - x2) ** 2).sum(dim=(1,2))
+        with torch.no_grad():
+            # remove outliers e.g. where LSTM went to different paths.
+            limit = 2.0 * torch.median(sumsq_diff)
+        indexes = torch.where(sumsq_diff < limit)
+        rms = sumsq_diff[indexes].sum().sqrt()
```
pzelasko, line 71: A small epsilon added here would ensure numerical stability when the diff is going to zero

pzelasko:
(We are discussing it with Yiwen so I put the comments here not to forget that discussion)

snowfall/models/normalize.py
danpovey: I don't think it's possible for it to go to zero.  At least half the indexes are present (since the cutoff is twice the median), and the sum is a sum of squares of thousands of Gaussian-distributed values, which is incredibly unlikely to be zero.

snowfall/models/normalize.py
danpovey: Agreed.

danpovey: ... should be possible to test `if self.training` for this, I believe nn.Module gives you that.

PR #71, opened 2021-01-07
[WIP] Initially working multi-gpu training

pzelasko:
I've only seen ~300 steps but the model seems to be converging alright, and both GPUs are close to 100% load. This probably needs further work to make sure the checkpoints work ok and the dev data scores are aggregated from all nodes. It also needs a few `if`s to handle both single-gpu and multi-gpu correctly. 

You can run this like:

```
# terminal 1
python mmi_bigram_train.py --world-size 2 --rank 0
```
```
# terminal 2
python mmi_bigram_train.py --world-size 2 --rank 1
```

You probably expected that to work already, but it's nice to see K2 running on multi-gpu :)

PR #95, opened 2021-02-09
Sync snowfall with Lhotse's dataset/sampler refactoring PR

pzelasko:
Reflects the changes from https://github.com/lhotse-speech/lhotse/pull/194

This new version allows working with distributed training out-of-the-box. Once it is merged, I will rebase #71 on that and finish multi-GPU training support.

One thing I am concerned about is the number of files with duplicate code. It becomes quite labor-intensive to propagate these kinds of changes, and I am not sure if I'm doing everything OK as I can't run all of these (🙏🏻 static checker tool helps 🙏🏻). It could be good to slowly converge to some common code structure that can be shared and imported (e.g. a `Trainer` class or something like that)?

### DIFF
--- a/egs/aishell/asr/simple_v1/ctc_decode.py
+++ b/egs/aishell/asr/simple_v1/ctc_decode.py
@@ -3,26 +3,24 @@
 # Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey, Haowen Qiu)
 # Apache 2.0
 
+import k2
 import logging
 import os
+import torch
+from k2 import Fsa, SymbolTable
+from kaldialign import edit_distance
 from pathlib import Path
 from typing import List
 from typing import Optional
 from typing import Union
 
-import k2
-import torch
-from k2 import Fsa, SymbolTable
-from kaldialign import edit_distance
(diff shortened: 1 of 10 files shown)


danpovey:
How do you run the static checker tool?
I'm OK with leaving snowfall fairly unstructured.   As far as I am concerned it is just a repo where we can iterate rapidly on experiments, and not a serious thing that we are "releasing"; we can do a cleaned-up repo later once we have an idea what kinds of models we'll be using.

pzelasko:
I use the one that is integrated into PyCharm and IMO it is the best tool available. I believe the next best thing is [mypy](https://github.com/python/mypy), and it has integrations with multiple editors/IDEs. Years ago when these things were not available I used [flake8](https://flake8.pycqa.org/en/latest/) it says it's "style enforcement" but it also has some static checks (and a rich plugin ecosystem I think). In case you are using VSCode, perhaps [pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) is interesting (it's exclusive to VSCode though).

pzelasko:
OK this works - merging.

PR #98, opened 2021-02-11
aishell recipe : fix typo && sync with latest K2

pingfengluo:
aishell recipe : 
1)  fix typo 
2)  sync with latest K2
3) add mmi_bigram RESULT


danpovey:
Thanks!!   Merging.  If there are any problems we'll address it later.

Issue #97, opened 2021-02-10
Auxiliary task (different type)


danpovey:
This is an auxiliary task for use with MMI training, as an alternative to the "attention" auxiliary task, that will be easier to do.  It is quite similar to the xent regularization in Kaldi.

From `num.get_arc_post().exp()` we can get the arc-level posteriors.  We can then construct a tensor `num_post_mat` of alignment posteriors of shape (seqframe_idx_dim, nnet_output_dim), where seqframe_idx_dim == dense_fsa_vec.shape_dim0() and nnet_output_dim is the num-phones plus one.  We use some kind of index_add_() construct to add in the posteriors to it; this will require `num.label` and `num.seqframe_idx`, assuming we gave `seqframe_idx_name='seqframe_idx'` to intersect_dense() for num.   Or use a torch sparse matrix.  Remember we have to eliminate elements where the labels are minus one, first, to avoid invalid indexing with `label < 0`.

Next, we need to give the neural network an auxiliary output layer that is structurally the same as the regular output, but with different parameters (it might be best to just have it return a single output that we divide into two pieces?).  The main extra part of the objective function would be  `xent_like = (xent_output * num_post_mat).sum()`, and note that xent_output needs to have been through a log-softmax or this is meaningless, i.e. it needs to be normalized logprobs.  From that I want to subtract the (appropriately scaled) negentropy of the distribution, to avoid the network being encouraged to put all the probability on one class.   Something like:  `phone_counts = num_post_mat.sum(0); xent_like = (nnet_output.log() * num_post_mat).sum() - (phone_counts * torch.log(phone_counts/phone_counts.sum() + 1.0e-20)).sum()`

danpovey:
@zhu-han if you could work on this it would be great.

zhu-han:
Working on it.

danpovey:
Great!

danpovey:
My understanding is that we tried this and it did not work.. closing the issue.

PR #100, opened 2021-02-16
[WIP] Auxiliary task similar to xent regularization

zhu-han:
Implement an auxiliary task described in https://github.com/k2-fsa/snowfall/issues/97#issue-805286306.

Tried two types of implementation. 

1. In addition to the MMI loss, we have two losses `xent_loss` and `smooth_loss`.
```
xent_loss = - (xent_output * num_post_mat).sum()
```
```
smooth_loss = (phone_counts * torch.log(phone_counts / phone_counts.sum() + 1.0e-20)).sum()
```
In this case, the regular output layer has three loss: MMI loss, `xent_loss` (since `num_post_mat` is from regular output layer) and `smooth_loss`.  And the auxiliary output layer has one loss: `xent_loss`.

2. Use `num_post_mat.detach()` to cut gradient from `xent_loss` to regular output layer. In this case,  I think `smooth_loss` may not be necessary. Therefore, the regular output layer has one loss: MMI loss. And the auxiliary output layer also has one loss: `xent_loss`.

Experiments are conducted with tdnn-lstm model. Haven't got a significant improvement yet.

- The baseline model is trained with MMI loss only. The model was trained three times. WERs are `10.39%`, `10.38%` and `10.11%`.

- Type 1 implementation results:

Configuration | WER
-- | --
xent_scale = 0.005   smooth_scale = 0.005 | 10.33%
xent_scale = 0.005 smooth_scale = 0.01 | 10.38%
xent_scale = 0.005 smooth_scale = 0.05 | 10.44%
xent_scale = 0.01 smooth_scale = 0.005 | 10.03%
xent_scale = 0.01 smooth_scale = 0.01 | 10.35%
xent_scale = 0.01 smooth_scale = 0.05 | 10.19%
xent_scale = 0.05 smooth_scale = 0.005 | 10.97%
xent_scale = 0.05 smooth_scale = 0.01 | 10.15%
xent_scale = 0.05 smooth_scale = 0.05 | 10.24%
xent_scale = 0.1 smooth_scale = 0.0 | 12.32%
xent_scale = 0.1 smooth_scale = 0.1 | 10.52%
xent_scale = 0.1 smooth_scale = 0.5 | 11.63%
xent_scale = 0.1 smooth_scale = 1.0 | 13.06%
xent_scale = 0.5 smooth_scale = 0.1 | 11.87%
xent_scale = 0.5 smooth_scale = 0.5 | 12.78%

, `xent_scale` and `smooth_scale` are loss scale of `xent_loss` and `smooth_loss` respectively.

- Type 2 implementation results:

Configuration | WER
-- | --
xent_scale = 0.005 | 10.14%
xent_scale = 0.01 | 10.45%
xent_scale = 0.05 | 10.13%
xent_scale = 0.1 | 10.23%
xent_scale = 0.3 | 10.81%
xent_scale = 0.5 | 10.32%
xent_scale = 1.0 | 11.87%

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_xent_train.py
+++ b/egs/librispeech/asr/simple_v1/mmi_xent_train.py
@@ -1,0 +1,734 @@
+#!/usr/bin/env python3
+# Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey
+#                                                   Haowen Qiu
+#                                                   Fangjun Kuang)
+# Apache 2.0
+
+import k2
+import logging
+import math
+import numpy as np
+import os
+import argparse
+import sys
+import torch
+import torch.optim as optim
+from datetime import datetime
+from pathlib import Path
+from torch import nn
+from torch.nn.utils import clip_grad_value_
+from torch.utils.tensorboard import SummaryWriter
+from typing import Dict, Optional, Tuple
+
+from lhotse import CutSet
+from lhotse.dataset import CutConcatenate, CutMix, K2SpeechRecognitionDataset, SingleCutSampler
+from lhotse.utils import fix_random_seed
+from snowfall.common import load_checkpoint, save_checkpoint
+from snowfall.common import save_training_info
+from snowfall.common import setup_logger
+from snowfall.common import describe
+from snowfall.models import AcousticModel
+from snowfall.models.tdnn_lstm import TdnnLstm1c
+from snowfall.training.diagnostics import measure_gradient_norms, optim_step_and_measure_param_change
+from snowfall.training.mmi_graph import MmiTrainingGraphCompiler
+from snowfall.training.mmi_graph import create_bigram_phone_lm
+from snowfall.training.mmi_graph import get_phone_symbols
+
+
+def get_tot_objf_and_num_frames(tot_scores: torch.Tensor,
+                                frames_per_seq: torch.Tensor
+                                ) -> Tuple[float, int, int]:
+    ''' Figures out the total score(log-prob) over all successful supervision segments
+    (i.e. those for which the total score wasn't -infinity), and the corresponding
+    number of frames of neural net output
+         Args:
+            tot_scores: a Torch tensor of shape (num_segments,) containing total scores
+                       from forward-backward
+        frames_per_seq: a Torch tensor of shape (num_segments,) containing the number of
+                       frames for each segment
+        Returns:
+             Returns a tuple of 3 scalar tensors:  (tot_score, ok_frames, all_frames)
+        where ok_frames is the frames for successful (finite) segments, and
+       all_frames is the frames for all segments (finite or not).
+    '''
+    mask = torch.ne(tot_scores, -math.inf)
+    # finite_indexes is a tensor containing successful segment indexes, e.g.
+    # [ 0 1 3 4 5 ]
+    finite_indexes = torch.nonzero(mask).squeeze(1)
+    if False:
+        bad_indexes = torch.nonzero(~mask).squeeze(1)
+        if bad_indexes.shape[0] > 0:
+            print("Bad indexes: ", bad_indexes, ", bad lengths: ",
+                  frames_per_seq[bad_indexes], " vs. max length ",
+                  torch.max(frames_per_seq), ", avg ",
+                  (torch.sum(frames_per_seq) / frames_per_seq.numel()))
+    # print("finite_indexes = ", finite_indexes, ", tot_scores = ", tot_scores)
+    ok_frames = frames_per_seq[finite_indexes].sum()
+    all_frames = frames_per_seq.sum()
+    return (tot_scores[finite_indexes].sum(), ok_frames, all_frames)
+
+
+def get_objf(batch: Dict,
+             model: AcousticModel,
+             P: k2.Fsa,
+             device: torch.device,
+             graph_compiler: MmiTrainingGraphCompiler,
+             is_training: bool,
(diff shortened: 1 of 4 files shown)


danpovey:
Mm, that's a shame.
Could you try the type 2 loss with
xent_scale = 0.05
and the type 1 loss with
xent_scale = 0.05, smooth_scale = 0.05
?

zhu-han:
Will try it.

danpovey:
Also please make sure that the `num_post_mat.sum(0)` is mostly ones with a few zeros (where the final frames are).
Just in case there is some bug I didn't spot.

zhu-han:
I think you means `num_post_mat.sum(1)`.  It is mostly ones with a few zeros as you said.

zhu-han:
Added type 2 `xent_scale = 0.05`  result and type 1 `xent_scale = 0.05, smooth_scale = 0.05` result to the above tables. With smaller scale, these two results are better. I could try smaller scales to see if we can get better performance.

BTW, I got 9.89% WER with the baseline model before. The latest result I got is 10.39%.  So the randomness of training could make it hard for us to ensure the effectiveness of this new task. I wonder should we make the training process deterministic?

danpovey:
I wonder if some other change in the code could have caused that difference in the baseline.  For example, I think we made a change in how the batching is done.  Perhaps try running the baseline once more?

zhu-han:
Sure, will try running it again.

egs/librispeech/asr/simple_v1/mmi_xent_train.py
```diff
@@ -0,0 +1,734 @@
...
+        cut_transforms=[
+            CutConcatenate(),
+            CutMix(
+                cuts=cuts_musan,
+                prob=0.5,
+                snr=(10, 20)
+            )
+        ]
+    )
+    train_sampler = SingleCutSampler(
```
pzelasko, line 574: @danpovey since he is using a `SingleCutSampler` and `CutConcatenate` a couple of lines before, I don't believe there are any significant changes to how the batches are formed w.r.t. the past experiments. For these to go into effect, you'd have to use the `BucketingSampler`.

danpovey: Oh OK. 
BTW, RE BucketingSampler: another concern I have is that if we use that, we'd have to do the same in test time because the model won't have "seen" data where there is a discontinuity.

zhu-han:
Updated https://github.com/k2-fsa/snowfall/pull/100#issue-573949092 with some new results. 
(1) I run the baseline models two times again, get WER 10.38% and 10.11%.
(2) With smaller `xent_scale` and `smooth_scale`, still can't get significant improvement. These differences may just come from randomness.

danpovey:
OK.  I feel we shouldn't give this up too readily because this is something that was consistently helpful in Kaldi.
Is it possible to have different versions of the final layer, e.g. the last lstm layer, for the regular vs. xent output?
I know this is a bit of a hassle, but I think it's worth doing the experiment.

zhu-han:
Sure. Could try it.

danpovey:
Can anyone make available for me one of the trained transformer models?
I want to have a look at the parameters..

zhu-han:
This is a transformer model trained with MMI: [epoch-19.pt.zip](https://github.com/k2-fsa/snowfall/files/5994395/epoch-19.pt.zip)

danpovey:
Thanks!

danpovey:
Empty zip file :-(

danpovey:
@zhu-han can you please check the zip file?

zhu-han:
Sorry, this one should be right: https://drive.google.com/file/d/1luEAQbapIB0mH000JaonkTsAZppu4DWa/view?usp=sharing

danpovey:
Thanks!

egs/librispeech/asr/simple_v1/mmi_xent_train.py
pzelasko: I think that cut concatenation has always been disabled in test time (dev/test Dataset/Dataloader)

danpovey:
@zhu-han would you mind sharing the epoch-1 and epoch-5 models  also?  I want to compare something..

zhu-han:
@danpovey Uploaded to https://drive.google.com/drive/folders/1V5L8WaLDDdg5_oJl-GV8CMpdl6rg9FIL?usp=sharing

pzelasko:
Regarding transformer: I noticed that "warmup steps" are 25000, isn't that a bit excessive? I think that would mean more than half of the training epochs are spent in the warmup mode... To my surprise Espresso/ESPnet recipes tend to stick to that value regardless of the size of the dataset, I wonder if this is not a relevant parameter, a magic value that "Just Works", or simply it's forgotten about?

zhu-han:
@pzelasko I think smaller warmup steps may be better given current setup. I haven't tuned this yet, could try it later.

zhu-han:
@danpovey I tried more experiments, such as the separate last lstm layer sturcture you mentioned before,  and a different training strategy: training the regular task for some epoch and then add the xent task for the remaining epochs. 

If we train the models for 20 epochs and decode after each epoch, we could find the mmi training without the xent task is consistently better than the one with xent task. The baseline mmi models could get WER around 9.8% between 12 ~ 20 epochs. However, with the xent task, WER is rarely below 10%.

I suspect the reason is the posteriors get from the regular output is less accurate than the one used in kaldi chain models, since no explicit frame level supervision is provided in snowfall's mmi training.

PR #109, opened 2021-02-23
Fix start_frame after subsampling in transformers

zhu-han:


danpovey:
Thanks!
Does this affect results?

zhu-han:
It affect a little but I think could be ignored, since it will affect mask value for only one frame.

PR #111, opened 2021-02-26
Saving state_dict of optimizer and lr_scheduler

zhu-han:


danpovey:
OK.  Are you confident that this is all correct?  I don't have enough familiarity with PyTorch optimizers to know.

danpovey:
.. thanks!!

zhu-han:
Yes, I think this is correct. I checked the learning rate and loss curve, training from an intermediate epoch is consistent with training from start.

danpovey:
Fantastic, thanks!  Merging.

PR #112, opened 2021-02-27
Fix subsampling with concatenation

zhu-han:


danpovey:
Do you think I should merge this now?

zhu-han:
I think we should wait @pzelasko and see if it could work with possible fix in lhotse, just in case there are any issues I missed.

pzelasko:
I think it's okay to merge and if I find anything else, I'll follow up

PR #113, opened 2021-03-02
[WIP] MMI graph compilation inside dataloader

pzelasko:
This code works only with DataLoader num_workers=0, and when we comment out the assertion `assert num.requires_grad == is_training`.

As to `num_workers`, when we set it to 1 or greater, it will spawn subprocesses and require the objects passed between to be pickle-able. I think that Fsa/FsaVec has no pickling support at the moment which is why the subprocess crashes (it's actually frustrating because you can't easily attach to it with GDB).

As to `num.requires_grad`, I am not sure if it needs to be True - the training seems to work fine when it's False? It's not like we're updating the numerator FSA parameters, since it gets deleted soon after the batch is processed anyway. In any case, the line where I set it to True actually does nothing - I'll get rid of it before we merge this (eventually).

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
@@ -113,11 +113,9 @@ def get_objf(batch: Dict,
     # nnet_output is [N, C, T]
     nnet_output = nnet_output.permute(0, 2, 1)  # now nnet_output is [N, T, C]
 
+    num, den = batch['mmi_numerator'], batch['mmi_denominator']
     if is_training:
-        num, den = graph_compiler.compile(texts, P)
-    else:
-        with torch.no_grad():
-            num, den = graph_compiler.compile(texts, P)
+        num.requires_grad = True
 
     assert num.requires_grad == is_training
     assert den.requires_grad is False
@@ -427,7 +425,9 @@ def main():
     graph_compiler = MmiTrainingGraphCompiler(
         L_inv=L_inv,
         phones=phone_symbol_table,
-        words=word_symbol_table
+        words=word_symbol_table,
+        # No need for cache since we'll compile graphs in DataLoader processes
+        use_cache=False
     )
     phone_ids = get_phone_symbols(phone_symbol_table)
     P = create_bigram_phone_lm(phone_ids)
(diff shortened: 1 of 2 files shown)


danpovey:
I think the issue with requires_grad is that we train the bigram language model by means of the numerator graph, so a grad does get propagated there.  That may actually make this approach non-workable, because it means that there is a connection between the model's P matrix and the num graph, and if the data-prep process is independent of the process that contains the model there may be a problem.  (Might not exist for certain other training mechanisms though.)

pzelasko:
I might be getting it wrong, but:
- `requires_grad=False` only means that you won't be storing the gradient information for a given tensor/object, but it can still be backprop'd through so that the preceding parameters get their gradient
- the bigram LM in `P.scores` is a separate entity from the numerator graph, that is attached to the model and gets normally updated (per https://github.com/k2-fsa/snowfall/pull/113/files#diff-f7ad18b0ff25da80362b606e8e147895f23b943ad66aea4e54aa8b17bc924d6bR527)

So I think if we're able to work around the pickling issues, it should work OK.

danpovey:
No, you can't backprop through it if it was forward prop'd with requires_grad=False.
The problem is that the connection between P.scores and the numerator is part of the graph that PyTorch creates, it's not really "movable", it wouldn't survive pickling.

pzelasko:
How about the following snippet? I think it shows the reverse.

```python
>>> import torch
>>> from torch import nn
>>> x = torch.randn(32, 100)
>>> l1 = nn.Linear(100, 200)
>>> l2 = nn.Linear(200, 10)
>>> l2.requires_grad_(False)
Linear(in_features=200, out_features=10, bias=True)
>>> l1.weight.requires_grad
True
>>> l2.weight.requires_grad
False
>>>
>>> z = l2(l1(x))
>>> z.sum().backward()
>>> l1.weight.grad
tensor([[-3.9523e-01,  1.0352e+00,  1.0160e-01,  ...,  3.4121e-02,
         -4.9239e-01, -3.2230e-01],
        [-7.5320e-02,  1.9729e-01,  1.9362e-02,  ...,  6.5026e-03,
         -9.3837e-02, -6.1422e-02],
        [ 5.9474e-03, -1.5578e-02, -1.5289e-03,  ..., -5.1346e-04,
          7.4096e-03,  4.8500e-03],
        ...,
        [ 3.4034e-01, -8.9147e-01, -8.7489e-02,  ..., -2.9383e-02,
          4.2401e-01,  2.7754e-01],
        [ 8.4615e-02, -2.2163e-01, -2.1751e-02,  ..., -7.3051e-03,
          1.0542e-01,  6.9002e-02],
        [ 1.7086e-01, -4.4754e-01, -4.3921e-02,  ..., -1.4751e-02,
          2.1286e-01,  1.3933e-01]])
>>> l2.weight.grad
>>>
```

danpovey:
think of it in terms of the computation tree.
the matrix product thats.computed.will have.requires grad.true.

pzelasko:
OK, I went back through the MmiTrainingGraphCompiler code and I think I understand now. I previously missed the point that P LM is used not only for the dens, but it also affects the nums score. In that case yeah, we can't do the composition in a subprocess, as it would have kept using the initial, un-updated version of `P`. (I'm spelling it out to make sure I got you right)

danpovey:
Yes.

pzelasko:
I am closing it since it's a dead-end at this time.

danpovey:
ok.

PR #115, opened 2021-03-02
Fix concatenation in transformer

zhu-han:


snowfall/models/transformer.py
```diff
@@ -585,23 +585,10 @@ def encoder_padding_mask(supervisions: Optional[Dict] = None) -> Optional[Tensor
...
-    indices = torch.argsort(supervision_segments[:, 1], descending=False)
-    supervision_segments = supervision_segments[indices]
-
-    indices = torch.argsort(supervision_segments[:, 0], descending=False)
-    supervision_segments = supervision_segments[indices]
-
-    previous_idx = -1
-    cur_length = 0
-    lengths = []
+    lengths = [0 for _ in range(int(max(supervision_segments[:, 0])) + 1)]
```
pzelasko, line 588: Is there a reason not to make it an empty list and append each item in the loop below?

snowfall/models/transformer.py
```diff
@@ -645,23 +632,9 @@ def get_normal_transcripts(supervision: Dict, words: k2.SymbolTable, oov: str =
...
-
-    previous_idx = -1
-    batch_text = []
-    for sequence_idx, text in zip(supervision_segments[0], texts_ids):
-        if sequence_idx == previous_idx:
-            batch_text[-1] = batch_text[-1] + text
-        else:
-            batch_text.append(text)
-        previous_idx = sequence_idx
+    batch_text = [[] for _ in range(int(max(supervision['sequence_idx'])) + 1)]
```
pzelasko, line 635: Same question as above

snowfall/models/transformer.py
zhu-han: If make a empty list and append in the loop, we have to decide whether it is the first sentence of this 'sequence_idx'. If true, we should append it. if not, we should update the last value in the list. I think initialize them first could keep the loop simple. But the other way is also good to me.

pzelasko: OK got it - I didn't notice that `sequence_idx` doesn't have to be sorted ascendingly.

zhu-han:
This fix could affect results of transformers. Will update the new results soon.

pzelasko:
Cool! I can run the experiment with bucketing + concatenation now too.

pzelasko:
Looks good to merge!

PR #114, opened 2021-03-02
transformer train/decode scripts for aishell

hegc:

Issue #75, opened 2021-01-17
Debug possible memory leaks


danpovey:
We need to debug possible memory leaks in k2, in master and also the arc_scores branch from my repo (see a PR on k2 master).
(I'm pretty sure arc_scores branch leaks; not 100% sure about master).

(1) Monitor memory usage from nvidia-smi and/or adding statements to the training script, and verify whether memory seems to increase.

(2) Look for available diagnostics from torch; I believe it can print out info about blocks allocated.

 or

(3) [somewhat alternative to (2)]    Add print statements to the constructor and destructor(/equivalents) of the Fsa object to check that the same number are constructed/destroyed on each iter.  If those are not the same: add similar print statements to the Function objects in autograd.py and see which of them are not destroyed.   I suspect it's a version of a previous issue where we had reference cycles between Fsa objects and Function objects used in backprop.

danpovey:
BTW, these leaks may actually be in the decoding script.  Should probably check both train and decode (that memory usage doesn't systematically increase with iteration.)

csukuangfj:
I am running `mmi_bigram_train.py`. There is no OOM after processing 1850 batches. `nvidia-smi` shows the used RAM is about 19785 MB.

danpovey:
The key thing is whether ti systematically increases over time.

` torch.cuda.get_device_properties(0).total_memory` may help

csukuangfj:
`torch.cuda.get_device_properties(0).total_memory` returns the property of the device, e.g., its capacity, which is a constant.

---

From https://pytorch.org/docs/stable/cuda.html#torch.cuda.memory_allocated, I am using

```python
torch.cuda.memory_allocated(0) / 1024.  # KB
```

The result is shown below. The allocated memory is increased monotonically. It is increased by about 500 KB every 10 batches.

---

The branch `arc_scores` is used.



![Screen Shot 2021-01-17 at 18 29 50](https://user-images.githubusercontent.com/5284924/104837822-13a76700-58f2-11eb-9229-4d44207772ca.png)

csukuangfj:
The figure matches the output `Allocated memory` from  `torch.cuda.memory_summary(0)`.

danpovey:
OK.  Try adding print statements to the Fsa object init/destructor (there
is a way, I forget).. to see if they are properly released.

csukuangfj:
```python
def __del__(self):
  print('inside Fsa destructor')
```

I think the above destructor will work.

[...]

PR #117, opened 2021-03-05
Fix OOM issue in AIShell recipe 

danpovey:


danpovey:
Relates to #114 and #75

danpovey:
This (reduction of beam from 10 to 6 in AIShell recipe) together with some code changes I just committed in k2, seems to resolve the OOM issue.

@csukuangfj there are some changes in mmi_graph.py which are from you.   Would you mind verifying that this is OK to merge?

csukuangfj:
> @csukuangfj there are some changes in mmi_graph.py which are from you. Would you mind verifying that this is OK to merge?


Sure, will run MMI training for AIShell with this pullrequest.

danpovey:
@csukuangfj  I mainly meant looking at the code or letting me know that this was intended for merge.
I am running the  AIShell recipe.
(I need to check that librispeech does not crash though-- no GPUs left).

csukuangfj:
OK, looking into it. I tested it before when I made the changes and the `mmi_bigram_train.py` got a similar loss as 
the script before the changes.

egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
```diff
@@ -424,14 +424,19 @@ def main():
             L_inv = k2.arc_sort(L.invert_())
             torch.save(L_inv.as_dict(), lang_dir / 'Linv.pt')
 
+    device_id = 0
```
csukuangfj, line 427: Line 274
```python
assert P.is_cpu
```
has to be deleted.

csukuangfj:
The changes look good to me.

danpovey:
Added AIShell results.
Note, the CER  is 6.94%, while the best Kaldi number I see in egs/aishell/s5 is around 7.4%.  So if I am not mistaken we may have passed Kaldi's performance on this particular dataset.

danpovey:
And many thanks to @heguocai for his contribution!

Issue #108, opened 2021-02-23
Random sampling with concatenation vs bucketing


pzelasko:
FYI

I ran the transformer training with mmi + attention with different batch sampling settings to check which sampler gives better WER. The results are from the averaged model using the last 5 checkpoints (15-20 epoch). The trends using the single best checkpoint results are similar.

**Default: SingleCutSampler with CutConcatenate, warmup 25k, max-frames 50000**
```2021-02-22 10:17:51,487 INFO [mmi_att_transformer_decode.py:300] %WER 8.15% [4285 / 52576, 674 ins, 377 del, 3234 sub ]```

**BucketingSampler, warmup 25k, max-frames 50000**
```2021-02-19 09:43:08,320 INFO [mmi_att_transformer_decode.py:300] %WER 8.30% [4362 / 52576, 710 ins, 365 del, 3287 sub ]```

**BucketingSampler, warmup 10k, max-frames 70000**
```2021-02-18 22:57:43,271 INFO [mmi_att_transformer_decode.py:300] %WER 8.31% [4370 / 52576, 697 ins, 362 del, 3311 sub ]```

**BucketingSampler, warmup 1k, max-frames 70000**
```2021-02-19 16:17:52,686 INFO [mmi_att_transformer_decode.py:300] %WER 8.32% [4372 / 52576, 709 ins, 376 del, 3287 sub ]```

It seems that with bucketing we're getting a small degradation, but we are able to train faster since we can set max-frames to a larger number. I think this could be because the cut concatenation works like a regularizer for the network, i.e. forces it not to "pay attention" to the utterances that do not matter for the current recognition. 

I wanted to check whether combining bucketing with concatenation would help, but I encountered an error inside the transformer - the encoder mask and encoder activation shapes are differing by 1 (e.g. 253 and 254). I don't know yet whether that is due to an issue in Lhotse or an issue in transformer code; but I don't know when I'll have enough time to debug it, so I wanted to share these findings in case somebody wants to pick it up or wonders which sampling worked better...

**Comment by danpovey** [comment]
Thanks!!

**Comment by zhu-han** [comment]
I find there is a bug about `start_frame` after subsampling in transformers, which could lead to a wrong mask. Fixed in https://github.com/k2-fsa/snowfall/pull/109.

**Comment by pzelasko** [comment]
Cool! In that case I'll re-attempt this.

**Comment by pzelasko** [comment]
@zhu-han I tried again with your fix, but I'm still getting the following error:

```
[...]

PR #118, opened 2021-03-05
Args for bucketing and concatenation, option to train on full librispeech

pzelasko:
In addition to the changes mentioned in #108, I am running the training on full libri to see what's the gap between our recipe and SOTA. I increased the model size (attention dim 256 -> 512 and nheads 4 -> 8) to resemble the ESPnet setup more closely. So far I've trained up until 6th epoch and the WER for averaged model on the 3 last epochs is 6.03%. It is improving each epoch so I will keep training and we'll see where we get.


pzelasko:
To recreate my bucketing + concatenation setup from #108, run: 

`python mmi_att_transformer_train.py --bucketing_sampler y --concatenate-cuts y --duration-factor 2`

danpovey:
Great!!

PR #119, opened 2021-03-08
Support Conformer

zhu-han:
For librispeech, conformer achives 6.84% WER, while transformer is 7.72%.
This PR also reduce warmup steps from 25k to 5k. After this modification, we could get comparable results in 10 epochs.

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -26,6 +26,7 @@ from snowfall.common import setup_logger
 from snowfall.decoding.graph import compile_LG
 from snowfall.models import AcousticModel
 from snowfall.models.transformer import Transformer
+from snowfall.models.conformer import Conformer
 from snowfall.training.ctc_graph import build_ctc_topo
 from snowfall.training.mmi_graph import create_bigram_phone_lm
 from snowfall.training.mmi_graph import get_phone_symbols
@@ -157,9 +158,15 @@ def print_transition_probabilities(P: k2.Fsa, phone_symbol_table: SymbolTable,
 def get_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument(
+        '--model-type',
+        type=str,
+        default="conformer",
+        choices=["transformer", "conformer"],
+        help="Model type.")
+    parser.add_argument(
         '--epoch',
         type=int,
-        default=20,
+        default=10,
         help="Decoding epoch.")
     parser.add_argument(
         '--max-frames',
@@ -193,12 +200,13 @@ def get_parser():
 def main():
     args = get_parser().parse_args()
 
+    model_type = args.model_type
     epoch = args.epoch
     max_frames = args.max_frames
     avg = args.avg
     att_rate = args.att_rate
 
-    exp_dir = Path('exp-transformer-noam-mmi-att-musan')
+    exp_dir = exp_dir = Path('exp-' + model_type + '-noam-mmi-att-musan')
     setup_logger('{}/log/log-decode'.format(exp_dir), log_level='debug')
 
     # load L, G, symbol_table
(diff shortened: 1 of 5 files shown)


danpovey:
Wow, great work!!

danpovey:
Will merge if no-one has objections in the next half hour...

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -193,12 +200,13 @@ def get_parser():
...
     args = get_parser().parse_args()
 
+    model_type = args.model_type
     epoch = args.epoch
     max_frames = args.max_frames
     avg = args.avg
     att_rate = args.att_rate
 
-    exp_dir = Path('exp-transformer-noam-mmi-att-musan')
+    exp_dir = exp_dir = Path('exp-' + model_type + '-noam-mmi-att-musan')
```
csukuangfj, line 209: There is an extra `exp_dir`.

snowfall/models/conformer.py
```diff
@@ -0,0 +1,711 @@
...
+            k = nn.functional.linear(key, _w, _b)
+
+            # This is inline in_proj function with in_proj_weight and in_proj_bias
+            _b = in_proj_bias
+            _start = embed_dim * 2
+            _end = None
+            _w = in_proj_weight[_start:, :]
+            if _b is not None:
+                _b = _b[_start:]
+            v = nn.functional.linear(value, _w, _b)
```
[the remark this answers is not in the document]
pzelasko: Oh, unless this is an unfolded loop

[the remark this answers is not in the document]
pzelasko: Just in case - the main branch (before else) also has a duplicated block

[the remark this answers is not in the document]
pzelasko: OK I see now that different variables are being computed `q`, `k`, and `v` - sorry for the comment spam

danpovey:
@zhu-han is there any chance you could either show some training diagnostics or upload a tensorboard log like the previous entry in the RESULTS file?

zhu-han:
Sure.  Latest mmi Transformer training log is :
[log-train-2021-03-07-02-27-09.txt](https://github.com/k2-fsa/snowfall/files/6116390/log-train-2021-03-07-02-27-09.txt)
And mmi Conformer training log is:
[log-train-2021-03-07-02-26-29.txt](https://github.com/k2-fsa/snowfall/files/6116417/log-train-2021-03-07-02-26-29.txt)
These are actually 20 epochs training log. But the best result are obtained with 10 epochs for both transformer and conformer.
Should I make another PR to update the RESULTS file?

danpovey:
Would be good if you could update the RESULTS file with these links.

snowfall/models/conformer.py
```diff
@@ -0,0 +1,711 @@
...
+        self.linear_pos = nn.Linear(embed_dim, embed_dim, bias=False)
+        # these two learnable bias are used in matrix c and matrix d
+        # as described in "Transformer-XL: Attentive Language Models Beyond a Fixed-Length Context" Section 3.3
+        self.pos_bias_u = nn.Parameter(torch.Tensor(num_heads, self.head_dim))
+        self.pos_bias_v = nn.Parameter(torch.Tensor(num_heads, self.head_dim))
+
+        self._reset_parameters()
+
+    def _reset_parameters(self) -> None:
+        nn.init.xavier_uniform_(self.in_proj.weight)
```
csukuangfj, line 341: `self.out_proj.weight` is left as default initialized.
Should we also initialize it here?

zhu-han: I‘m not sure. I refer to `MultiheadAttention`  in PyTorch (https://pytorch.org/docs/stable/_modules/torch/nn/modules/activation.html#MultiheadAttention) for this implementation. Don't know why they left `self.out_proj.weight` as default initialized.

snowfall/models/conformer.py
```diff
@@ -0,0 +1,711 @@
...
+        src = residual + self.dropout(self.conv_module(src))
+        if not self.normalize_before:
+            src = self.norm_conv(src)
+
+        # feed forward module
+        residual = src
+        if self.normalize_before:
+            src = self.norm_ff(src)
+        src = residual + self.ff_scale * self.dropout(self.feed_forward(src))
+        if not self.normalize_before:
```
csukuangfj, line 175: If `self.normalize_before` is False, then the code becomes
```python
src = self.norm_ff(src)
src = self.norm_final(src)
```

You see that it uses two  LayerNorms without anything in-between.

It is not a problem for the original paper since it always uses pre-LayerNorm,
i.e., with `self.normalize_before` always being True.

PR #124, opened 2021-03-10
Renaming LG -> HLG

### LINKED ISSUE #121: Issue in naming: LG vs HLG
I was explaining snowfall's approach to decoding to somebody and I realized that what we call "LG" might actually be "HLG", because the actual LG is composed with `ctc_topo` (our `H`); see e.g. in the following contexts:

https://github.com/k2-fsa/snowfall/blob/master/snowfall/decoding/graph.py#L66
https://github.com/k2-fsa/snowfall/blob/master/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py#L71

If you agree with that I will rename it.

pzelasko:
Resolves #121

### DIFF
--- a/egs/aishell/asr/simple_v1/ctc_decode.py
+++ b/egs/aishell/asr/simple_v1/ctc_decode.py
@@ -20,14 +20,14 @@ from snowfall.common import get_phone_symbols
 from snowfall.common import get_texts
 from snowfall.common import load_checkpoint
 from snowfall.common import setup_logger
-from snowfall.decoding.graph import compile_LG
+from snowfall.decoding.graph import compile_HLG
 from snowfall.models import AcousticModel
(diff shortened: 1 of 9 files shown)


pzelasko:
@danpovey @csukuangfj LMK if I can merge it.

csukuangfj:
Thanks! +2

xiaohui-zhang:
nit: calling it TLG (T: topo) might be better since the we don't call "ctc_topo" FST H.fst explicitly. Also it'll be less confusing in case later on we want to compare different label topologies where H.fst really means HMM.

danpovey:
@xiaohui-zhang  We spoke about this issue before; I felt H was better because not everyone has used Eesen so knows the T terminology.  Also I do regard the CTC topology as just a specific kind of H, rather than a special different thing.

PR #130, opened 2021-03-16
[WIP] Training with on-the-fly feature extraction

pzelasko:
This is how on-the-fly feature extraction would look like. The main differences are that:
- the cuts (speech and noise) are mixed and padded in the time domain, not fbank domain
- there is no lilcom compression
- [not in this PR yet] speed perturbation could be applied stochastically; instead of creating 3x larger epochs, we can have 3x more epochs and apply perturbation with a given probability

I think I will revert `'inputs'` key back to `'features'` in batches, and replace `max_frames` everywhere with `max_duration` (the relationship is simple - multiply duration by 100 and you have the num_frames). I'll also add option to `prepare.py` not to extract features.

Preliminary results are actually worse for on-the-fly variant, but I haven't figured out why yet. I'll have to inspect the batches more carefully to understand that.

**Baseline WER (pre-computed features + feat-domain mixing):**
```2021-03-12 22:46:33,175 INFO [mmi_att_transformer_decode.py:330] %WER 6.82% [3587 / 52576, 512 ins, 284 del, 2791 sub ]```

**On-the-fly WER:**
```2021-03-16 09:27:06,059 INFO [mmi_att_transformer_decode.py:330] %WER 7.01% [3684 / 52576, 512 ins, 284 del, 2888 sub ]```

### DIFF
--- a/egs/aishell/asr/simple_v1/ctc_decode.py
+++ b/egs/aishell/asr/simple_v1/ctc_decode.py
@@ -32,7 +32,7 @@ def decode(dataloader: torch.utils.data.DataLoader, model: AcousticModel,
     num_cuts = 0
     results = []  # a list of pair (ref_words, hyp_words)
     for batch_idx, batch in enumerate(dataloader):
-        feature = batch['features']
+        feature = batch['inputs']
         supervisions = batch['supervisions']
         supervision_segments = torch.stack(
             (supervisions['sequence_idx'],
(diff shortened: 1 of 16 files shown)


pzelasko:
One note: there is a useful function in `torchaudio` that's called `bg_iterator`, it starts a background thread that's reponsible for prefetching the batches from dataloader processes, it makes a significant difference for on-the-fly computation (without it there's 200-300ms delay, with it it's practically the same as the pre-computed variant).

danpovey:
Great!!
There is a small random variation in results.  That 0.1% absolute difference, I would not treat as reliable.

pzelasko:
OK - I made last changes for consistency, and I'm merging now. Please update your Lhotse to the latest upstream version after that.

pzelasko:
I'll still look into whether there are any differences between precomputed and on-the-fly batches, just later.

PR #134, opened 2021-03-19
Add dev-other and test-other to librispeech recipe

pzelasko:
Note: this is un-tested yet, I'll run it once I'm able to...

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -16,9 +16,9 @@ from pathlib import Path
 from typing import List
 from typing import Union
 
-from lhotse import CutSet
+from lhotse import CutSet, load_manifest
 from lhotse.dataset import K2SpeechRecognitionDataset, SingleCutSampler
-from snowfall.common import average_checkpoint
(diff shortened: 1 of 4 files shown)


danpovey:
Great!   I'll wait for you to test before merging..

pzelasko:
The servers where I have my recently-trained models are currently down; if you need this quickly then you might need to try it out yourself for now 😕

danpovey:
When I tested an already-trained model with this (trained with an older version of the code), I got 11% on test-clean where I should have got around 7%.  I wonder if there might have been any change that could be responsible for this, e.g. some code fix about the attention mask?
I don't think it's about this PR, I don't see anything that would affect this.

danpovey:
The WER regression wasn't to do with this, and this works otherwise, so merging.

pzelasko:
Interesting, I will check as soon as I have access to my models again.

pzelasko:
I checked out this branch and ran with lhotse before the latest bug-fixes; got WER 6.82% on test-clean, with a conformer model trained ~a week ago. Then I checked out latest lhotse and got the same number. Then, I updated K2 to the latest version and it's still the same number. I have no idea where the regression you reported comes from.

danpovey:
It's OK about the regression, I think it's a problem with a model I trained a very long time ago.

PR #136, opened 2021-03-23
Fix for latest k2 version

danpovey:


danpovey:
This PR is required so that the changes in https://github.com/k2-fsa/k2/pull/690 will not cause a crash.

PR #141, opened 2021-03-28
Modify how WERs are printed

danpovey:


danpovey:
I have tested this in a different branch.

PR #143, opened 2021-03-29
Add results for CTC conformer training with SpecAug

csukuangfj:
Closes #142 

---

- Tensorboard log <https://tensorboard.dev/experiment/s5iepd7PQBmzKOodcMH6Wg/>
- Training log [log-train-2021-03-29-18-17-44.txt](https://github.com/k2-fsa/snowfall/files/6225743/log-train-2021-03-29-18-17-44.txt)

PR #145, opened 2021-04-02
Support multi-GPU training with DDP from PyTorch.

csukuangfj:
The training time is approximately halved using two GPUs.

But it got stuck at the line
```python
cleanup_dist()
```
when the training was about to finish. So only checkpoints up to `epoch-8.pt` are available.

The decoding results with model averaging using the last 5 checkpoints are
```
2021-04-02 09:31:15,951 INFO [common.py:356] [test-clean] %WER 7.11% [3736 / 52576, 487 ins, 314 del, 2935 sub ]
2021-04-02 09:33:42,194 INFO [common.py:356] [test-other] %WER 18.50% [9686 / 52343, 1080 ins, 889 del, 7717 sub ]
```

The DDP training seems working.

---

After removing `cleanup_dist()` and testing it with a small amount of data, the training process is able to finish in a correct way.
The training process is started again and will run for 10 epochs. I will report the result here after it finishes.

---

Tensorboard logs: 
  - first try (2 GPUs): https://tensorboard.dev/experiment/fJrc3otkQ2WnXHP6pY0nrw/
  - second try (2 GPUs): https://tensorboard.dev/experiment/xY215fYATxyj8d78c5DlLw
  - third try (2 GPUs): https://tensorboard.dev/experiment/LdD8mokDTaWACV8Mbvh5CQ/
  - fourth try (3 GPUs): https://tensorboard.dev/experiment/3YrmNs2dSL6c5lELeg23Fg/

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
@@ -33,6 +33,7 @@ from snowfall.common import describe, str2bool
 from snowfall.common import load_checkpoint, save_checkpoint
 from snowfall.common import save_training_info
 from snowfall.common import setup_logger
+from snowfall.dist import cleanup_dist
 from snowfall.dist import setup_dist
 from snowfall.models import AcousticModel
 from snowfall.models.conformer import Conformer
@@ -367,7 +368,7 @@ def get_parser():
         '--num-epochs',
         type=int,
         default=10,
-        help='Number of traning epochs.')
+        help='Number of training epochs.')
     parser.add_argument(
         '--start-epoch',
         type=int,
@@ -747,9 +748,24 @@ def run(rank, world_size, args):
                            local_rank=rank)
 
     logging.warning('Done')
-    # CAUTION(fangjun): We have to comment out the following statement.
-    # Otherwise, the training seems never to terminate and it gets stuck here.
-    #  cleanup_dist()
+    torch.distributed.barrier()
+    # NOTE: The training process is very likely to hang at this point.
+    # If you press ctrl + c, your GPU memory will not be freed.
+    # To free you GPU memory, you can run:
+    #
+    #  $ ps aux | grep multi
+    #
+    # And it will print something like below:
+    #
+    # kuangfa+  430518 98.9  0.6 57074236 3425732 pts/21 Rl Apr02 639:01 /root/fangjun/py38/bin/python3 -c from multiprocessing.spawn
+    #
+    # You can kill the process manually by:
+    #
+    # $ kill -9 430518
+    #
+    # And you will see that your GPU is now not occupied anymore.
+    cleanup_dist()
+
 
 def main():
     args = get_parser().parse_args()
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
@@ -14,9 +14,11 @@ import numpy as np
 import os
 import sys
 import torch
+import torch.multiprocessing as mp
 from datetime import datetime
 from pathlib import Path
 from torch import nn
+from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.nn.utils import clip_grad_value_
 from torch.utils.tensorboard import SummaryWriter
 from typing import Dict, Optional, Tuple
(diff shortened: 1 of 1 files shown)


danpovey:
Cool!

pzelasko:
I think I have an idea what could be causing the hanging. It seems synchronization with NCCL requires using torch.distributed.barrier() before syncing tensors. I have it on my list to check it but if you are working on it atm maybe you want to try.

csukuangfj:
> . It seems synchronization with NCCL requires using torch.distributed.barrier() before syncing tensors.

Thanks, will try it. 

---

I've set num_epochs to 11 so that even if it hangs, we can still get `epoch-9.pt`. It indeed hung. Will try `torch.distributed.barrier()` before exiting the process.

---

Here are the WERs:

(1) averaging with checkpoints from the last 5 epochs
```
2021-04-02 20:50:11,390 INFO [common.py:356] [test-clean] %WER 7.00% [3678 / 52576, 493 ins, 290 del, 2895 sub ]
2021-04-02 20:52:20,198 INFO [common.py:356] [test-other] %WER 18.15% [9501 / 52343, 1058 ins, 942 del, 7501 sub ]
```

It is slightly worse than the results recorded in RESULTS.md:  `6.90` vs `7.00` and `17.89` vs `18.15`. 

Will report WERs per epoch later.

As for the training time, we can use the time points when learning rate data is recorded in the tensorboard log to infer the training time per epoch
- #143 https://tensorboard.dev/experiment/h3xiWY0oQ4WGd2dRgG8NWw/
   - About 52 minutes per epoch
  
- this pull-request (with 2 GPUs):  https://tensorboard.dev/experiment/LdD8mokDTaWACV8Mbvh5CQ
   - About 26 minutes per epoch
  
You can see that the training time is halved with 2 GPUs and the resulting WERs are comparable to that trained with single GPU.

pzelasko:
Cool. Probably we need to tune the learning rate to account for larger effective batch sizes.

As to distributed: I was recently using Lhotse in another project where the training loop is implemented with Pytorch Lightning, and multi-GPU training did never hang on this one. I did an initial scan of what they're doing in Lightning, and [that's how I found about the barrier thing](https://github.com/PyTorchLightning/pytorch-lightning/blob/217c12a4e7251681d3840b70db54e1f733b8d67c/pytorch_lightning/utilities/distributed.py#L120). Maybe there is other useful stuff to be gained from there.

csukuangfj:
> Maybe there is other useful stuff to be gained from there.

Thanks for the information. Will take a look into it.

egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
```diff
@@ -231,14 +234,14 @@ def train_one_epoch(dataloader: torch.utils.data.DataLoader,
                     tb_writer: SummaryWriter,
                     num_epochs: int,
                     global_batch_idx_train: int):
-    """One epoch training and validation.
+    '''One epoch training and validation.
```
pzelasko, line 237: I think it's a bad pattern to do some many formatting changes in a PR like this; it's very distracting from what's the actual contribution.

egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
```diff
@@ -585,7 +608,7 @@ def main():
         cuts_dev,
         max_duration=max_duration,
```
pzelasko, line 609: You have two options regarding the validation dataset:
1) distribute it to all nodes and synchronize dev loss
2) give 100% to all nodes and forego synchronization

By default, Lhotse's samplers are auto-distributing themselves when they detect distributed training. That means currently you are reporting the dev loss on a partial dev set. You need to either sync the loss across nodes or explicitly set rank=0 and world_size=1 to the dev sampler (to disable the auto-distribution mechanism).

It might be a good idea to explicitly set the rank and world size in train samplers too?

Also I'm not sure if this is a good (as in: least surprising) behaviour by Lhotse classes, I'm open to feedback.

pzelasko (commented):
I noticed that you are not synchronizing any of the metrics that are being tracked and reported: especially loss, dev loss and num frames seem like important candidates for accurate reporting.

egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py
csukuangfj: Ok, reverted.

csukuangfj:
There are two issues for this pull-request:

- You may have to kill the processes manually to free the GPU memory
- The valid objf is for a single node only.

Otherwise, it is ready to merge.

danpovey:
I mentioned somewhere else that the reason for the hang seems to be that the  sampler seems to give a different number of minibatches in different processes.  Merging anyway.

PR #147, opened 2021-04-05
Rescoring with 4-gram LM

csukuangfj:
~Support for rescoring with the whole lattice is not added yet, as k2 does not support removing epsilon self-loops at present.~

Here are the WERs (%) using LM rescoring. You can see that rescoring with LM reduces WER.

## Base line (MMI conformer training)
It uses `k2.shortest_path` for decoding
- test_clean: 6.80
- test_other: 18.03

## With LM rescoring
It uses `k2.random_paths` with various `num_paths`

Screen Shot 2021-04-05 at 1 23 36 PM

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -21,10 +21,12 @@ from lhotse.dataset import K2SpeechRecognitionDataset, SingleCutSampler
 from snowfall.common import average_checkpoint, store_transcripts
 from snowfall.common import find_first_disambig_symbol
 from snowfall.common import get_texts
-from snowfall.common import write_error_stats
 from snowfall.common import load_checkpoint
 from snowfall.common import setup_logger
+from snowfall.common import str2bool
+from snowfall.common import write_error_stats
 from snowfall.decoding.graph import compile_HLG
+from snowfall.decoding.lm_rescore import decode_with_lm_rescoring
 from snowfall.models import AcousticModel
 from snowfall.models.transformer import Transformer
 from snowfall.models.conformer import Conformer
@@ -34,7 +36,9 @@ from snowfall.training.mmi_graph import get_phone_symbols
 
 
 def decode(dataloader: torch.utils.data.DataLoader, model: AcousticModel,
-           device: Union[str, torch.device], HLG: Fsa, symbols: SymbolTable):
+        device: Union[str, torch.device], HLG: Fsa, G: Fsa, num_paths: int,
+        symbols: SymbolTable):
+    G = k2.arc_sort(G)  # It is a no-op if it is already sorted.
     tot_num_cuts = len(dataloader.dataset.cuts)
     num_cuts = 0
     results = []  # a list of pair (ref_words, hyp_words)
@@ -72,8 +76,10 @@ def decode(dataloader: torch.utils.data.DataLoader, model: AcousticModel,
         lattices = k2.intersect_dense_pruned(HLG, dense_fsa_vec, 20.0, 7.0, 30,
                                              10000)
 
-        # lattices = k2.intersect_dense(HLG, dense_fsa_vec, 10.0)
-        best_paths = k2.shortest_path(lattices, use_double_scores=True)
+        if G is None:
+            best_paths = k2.shortest_path(lattices, use_double_scores=True)
+        else:
+            best_paths = decode_with_lm_rescoring(lattices, G, num_paths=num_paths)
         assert best_paths.shape[0] == len(texts)
         hyps = get_texts(best_paths, indices)
         assert len(hyps) == len(texts)
@@ -178,7 +184,7 @@ def get_parser():
         '--avg',
         type=int,
         default=5,
-        help="Number of checkpionts to average. Automaticly select "
+        help="Number of checkpionts to average. Automatically select "
              "consecutive checkpoints before checkpoint specified by'--epoch'. ")
     parser.add_argument(
         '--att-rate',
@@ -195,6 +201,19 @@ def get_parser():
         type=int,
         default=256,
         help="Number of units in transformer attention layers.")
+    parser.add_argument(
+        '--use-lm-rescoring',
+        type=str2bool,
+        default=True,
+        help='When enabled, it uses LM for rescoring')
+    parser.add_argument(
+        '--num-paths',
+        type=int,
+        default=100,
+        help='Number of paths for rescoring using n-best list.' \
+             'If it is negative, then rescore without using n-best list.'\
+             'CAUTION: You have to reduce max_duration in case of error "Problem size is too large"'
+             )
     return parser
 
 
@@ -206,6 +225,12 @@ def main():
     max_duration = args.max_duration
     avg = args.avg
     att_rate = args.att_rate
+    use_lm_rescoring = args.use_lm_rescoring
+    num_paths = args.num_paths
+    if num_paths < 0:
+        num_paths = None
+    else:
+        assert num_paths > 1
 
     exp_dir = Path('exp-' + model_type + '-noam-mmi-att-musan-sa')
     setup_logger('{}/log/log-decode'.format(exp_dir), log_level='debug')
@@ -288,6 +313,40 @@ def main():
         d = torch.load(lang_dir / 'HLG.pt')
         HLG = k2.Fsa.from_dict(d)
 
+    if args.use_lm_rescoring:
+        logging.debug(f'Rescoring with num_paths: {num_paths}')
+        first_word_disambig_id = find_first_disambig_symbol(symbol_table)
+        logging.debug('Loading G_4_gram')
+        if not os.path.exists(lang_dir / 'G_4_gram.pt'):
+            with open(lang_dir / 'G_4_gram.fst.txt') as f:
+                G = k2.Fsa.from_openfst(f.read(), acceptor=False)
+                del G.aux_labels
+                # CAUTION(fangjun): The following line is crucial.
+                # Arcs entering the back-off state have label equal to #0.
+                # We have to change it to 0 here.
+                G.labels[G.labels >= first_word_disambig_id] = 0
+                G = k2.create_fsa_vec([G]).to(device)
+                G = k2.arc_sort(G)
+                torch.save(G.as_dict(), lang_dir / 'G_4_gram.pt')
+        else:
+            logging.debug('Loading pre-compiled G_4_gram')
+            d = torch.load(lang_dir / 'G_4_gram.pt')
+            G = k2.Fsa.from_dict(d).to(device)
+
+        if num_paths is None:
+            # TODO(fangjun): remove this
+            assert False, 'Rescoring with whole lattice is currently not supported'
+            # We are not going to use n-best list for rescoring.
+            # Add epsilon self-loops to G.
+            G = k2.add_epsilon_self_loops(G)
+            G = k2.arc_sort(G)
+            G = G.to(device)
+
+            logging.debug('Use whole lattice for rescoring')
+    else:
+        logging.debug('Decoding without LM rescoring')
+        G = None
+
     logging.debug("convert HLG to device")
     HLG = HLG.to(device)
     HLG.aux_labels = k2.ragged.remove_values_eq(HLG.aux_labels, 0)
@@ -309,11 +368,16 @@ def main():
         logging.debug("About to create test dataloader")
         test_dl = torch.utils.data.DataLoader(test, batch_size=None, sampler=sampler, num_workers=1)
 
+        if not hasattr(HLG, 'lm_scores'):
+            HLG.lm_scores = HLG.scores.clone()
+
         logging.debug("About to decode")
         results = decode(dataloader=test_dl,
                          model=model,
                          device=device,
                          HLG=HLG,
+                         G=G,
+                         num_paths=num_paths,
                          symbols=symbol_table)
 
         recog_path = exp_dir / f'recogs-{test_set}.txt'
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -21,8 +21,10 @@ from snowfall.common import get_texts
 from snowfall.common import write_error_stats
 from snowfall.common import load_checkpoint
 from snowfall.common import setup_logger
+from snowfall.common import str2bool
 from snowfall.data import LibriSpeechAsrDataModule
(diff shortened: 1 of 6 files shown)


snowfall/decoding/lm_rescore.py
```diff
@@ -0,0 +1,268 @@
...
+    # num_repeats.num_elements() == unique_word_seqs.num_elements()
+    #
+    # Since k2.ragged.unique_sequences will reorder paths within a seq,
+    # `new2old` is a 1-D torch.Tensor mapping from the output path index
+    # to the input path index.
+    # new2old.numel() == unique_word_seqs.num_elements()
+    unique_word_seqs, num_repeats, new2old = k2.ragged.unique_sequences(
+        word_seqs, need_num_repeats=True, need_new2old_indexes=True)
+
+    seq_to_path_shape = k2.ragged.get_layer(unique_word_seqs.shape(), 0)
```
pzelasko, line 127: side note: I think it would be very helpful to have a manual that describes these methods and how FSAs/Ragged/etc. can be manipulated, in a similar way to what numpy has (https://numpy.org/doc/stable/user/basics.indexing.html)

snowfall/decoding/lm_rescore.py
```diff
@@ -0,0 +1,268 @@
...
+                                  path_to_seq_map)
+
+    # Now compute lm_scores
+    b_to_a_map = torch.zeros_like(path_to_seq_map)
+    lm_path_lats = k2.intersect_device(G,
+                                       word_fsas_with_epsilon_loops,
+                                       b_to_a_map=b_to_a_map,
+                                       sorted_match_a=True)
+    lm_path_lats = k2.top_sort(k2.connect(lm_path_lats.to('cpu'))).to(device)
+    lm_scores = lm_path_lats.get_tot_scores(True, True)
```
pzelasko, line 153: what is the meaning of `(True, True)`? could these be keyword args instead?

snowfall/decoding/lm_rescore.py
```diff
@@ -0,0 +1,268 @@
...
+
+    tot_scores = am_scores + lm_scores
```
pzelasko, line 155: doesn't this way of rescoring remove the "graph scores" that come from e.g. multiple entries with distinct probabilities in the lexicon?

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -288,6 +312,40 @@ def main():
...
         HLG = k2.Fsa.from_dict(d)
 
+    if args.use_lm_rescoring:
+        logging.debug(f'Rescoring with num_paths: {num_paths}')
+        first_word_disambig_id = find_first_disambig_symbol(symbol_table)
+        logging.debug('Loading G_4_gram')
+        if not os.path.exists(lang_dir / 'G_4_gram.pt'):
+            with open(lang_dir / 'G_4_gram.fst.txt') as f:
+                G = k2.Fsa.from_openfst(f.read(), acceptor=False)
+                del G.aux_labels
```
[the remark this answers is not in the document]
csukuangfj: You can view an existing 3-gram G at https://github.com/csukuangfj/kaldilm/blob/master/G.svg
or by visiting https://github.com/csukuangfj/kaldilm

The resulting G is a transducer, not an acceptor. What we want is an acceptor, so we remove its `aux_labels` here.
If we don't remove it, its `aux_labels` will be propagated in later operations.

snowfall/decoding/lm_rescore.py
```diff
@@ -0,0 +1,268 @@
...
+
+    # The `scores` of every arc consists of `am_scores` and `lm_scores`
+    am_path_lats.scores = am_path_lats.scores - am_path_lats.lm_scores
+
+    am_scores = am_path_lats.get_tot_scores(True, True)
+
+    return am_scores
+
+
+@torch.no_grad()
```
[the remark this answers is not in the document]
csukuangfj: We are not using gradient in decoding.

`no_grad` can be removed in this case. It will not affect the result. But I believe it requires more memory as it
needs to save the graph information for backward.

I am not sure how gradient will be used in the decoding script. That's why I am using `no_grad` here.

snowfall/decoding/lm_rescore.py
```diff
@@ -0,0 +1,268 @@
...
+    assert hasattr(lats, 'aux_labels')
+    assert hasattr(lats, 'lm_scores')
+
+    assert G.shape == (1, None, None)
+    assert G.device == device
+    assert hasattr(G, 'aux_labels') is False
+
+    # First, extract `num_paths` paths for each sequence.
+    # paths is a k2.RaggedInt with axes [seq][path][arc_pos]
+    paths = k2.random_paths(lats, num_paths=num_paths, use_double_scores=True)
```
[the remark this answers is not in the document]
csukuangfj: `k2.random_paths` is a template function in C++ and it has instantiations for both `float` and `double`.

`use_doube_scores` means it will use `double`, instead of `float` in C++.
The precision of `double` is higher than `float`, so I am using `double` here.

You can use `use_double_scores=False`, but I am not sure how it will affect the result.

snowfall/decoding/lm_rescore.py
csukuangfj: I agree. 

We definitely need to improve the documentation from the Python side.

I guess the current documention in Python makes the most sense only to those who are familiar with k2 C++ APIs.

snowfall/decoding/lm_rescore.py
csukuangfj: The documentation about these two arguments can be found at
https://k2.readthedocs.io/en/latest/python_api/fsa.html#id2


`get_tot_scores` accepts two args:
-  use_double_scores:  This argument has the same meaning as its counterpart in `k2.random_paths`
- log_semiring: If it is true, it uses log semiring; otherwise, it uses tropical semiring.

> could these be keyword args instead?

Yes, I should have used keyword args here. But since both of them are True, so I just used positional arguments.

snowfall/decoding/lm_rescore.py
csukuangfj: I believe we are currently using a lexicon with log-prob 0 on all arcs, so the lexicon does not contribute probabilities to the final rescoring.

The current approach assumes that LM scores come exclusively from the language model.

pzelasko:
@csukuangfj thanks for clarifying!

snowfall/decoding/lm_rescore.py
csukuangfj: What I am currently doing is

- Get an n-best list.  (Note that the `best` in `n-best` is not the same one as in Kaldi. This list is sampled from a lattice)
- For each list (i.e., for each path), we remove the scores of very arc that come from the language model (i.e., HLG).
- Get the acoustic scores of each path
- Get the LM scores of each path
- Get the total scores of each path, with is am_scores + lm_scores
- The path with the largest total scores is the best path, which is used as the decoding output

csukuangfj: What I am currently doing is

- Get an n-best list.  (Note that the `best` in `n-best` is not the same one as in Kaldi. This list is sampled from a lattice)
- For each list (i.e., for each path), we remove the scores of very arc that come from the language model (i.e., HLG).
- Get the acoustic scores of each path
- Get the LM scores of each path
- Get the total scores of each path, with is am_scores + lm_scores
- The path with the largest total scores is the best path, which is used as the decoding output

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -288,6 +312,40 @@ def main():
...
+                G = k2.arc_sort(G)
+                torch.save(G.as_dict(), lang_dir / 'G_4_gram.pt')
+        else:
+            logging.debug('Loading pre-compiled G_4_gram')
+            d = torch.load(lang_dir / 'G_4_gram.pt')
+            G = k2.Fsa.from_dict(d).to(device)
+
+        if num_paths is None:
+            # TODO(fangjun): remove this
+            assert False, 'Rescoring with whole lattice is currently not supported'
```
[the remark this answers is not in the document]
csukuangfj: Will implement it once https://github.com/k2-fsa/k2/pull/705  is ready.

csukuangfj:
Here is the result for rescoring with the whole lattice.
(NOTE: It needs https://github.com/k2-fsa/k2/pull/705 to test this pull-request)

Screen Shot 2021-04-06 at 5 46 39 PM

You can see that rescoring with the whole lattice has a lower WER than that of n-best list, even if it has a smaller output beam.


---

I have to decrease the [output beam][1] size in `k2.intersect_dense_pruned` from  7 to `2` and `1` to avoid the following error.

```
[F] /ceph-fj/fangjun/open-source/s2/k2/k2/csrc/intersect.cu:1054:void k2::DeviceIntersector::ForwardSortedA() Problem size is too larg
e for this code: b_state_bits=18, key_bits=46, value_bits=18, value_max=262144, next_state_end=268477

```



[1]: https://github.com/k2-fsa/snowfall/blob/1255c3a287939701226d412ae1150add719dac3a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py#L75

danpovey:
Great!!
To-morrow I'll finish fixing the hash thing so that limit won't be a problem.

csukuangfj:
Here are updated WERs with a larger output beam using changes from https://github.com/k2-fsa/k2/pull/708

The WER for test-clean is decreased from 6.80% to 5.81%.

The WER is likely to further decrease with a larger output beam. We will see after Dan changes the implementation of hash in k2.

Screen Shot 2021-04-07 at 10 59 59 AM

danpovey:
Great!!  Will do it now...

snowfall/decoding/lm_rescore.py
```diff
@@ -217,10 +217,30 @@ def rescore_with_whole_lattice(lats: k2.Fsa,
...
+                                             sorted_match_a=True)
+    except RuntimeError as e:
+        print(f'Caught exception:\n{e}\n')
+        print(f'Number of FSAs: {inverted_lats.shape[0]}')
+        print('num_arcs before pruning: ',
+              inverted_lats_with_epsilon_loops.arcs.num_elements())
+
+        # NOTE(fangjun): The choice of the threshold 0.01 is arbitrary here
+        # to avoid OOM. We may need to fine tune it.
+        inverted_lats = k2.prune_on_arc_post(inverted_lats, 0.001, True)
```
csukuangfj, line 233: It needs https://github.com/k2-fsa/k2/pull/713

csukuangfj:
Here are the latest results.

Ready to review and merge.

### Average over last 5 epochs

#### LM rescoring with whole lattice

```
$ ./mmi_att_transformer_decode.py --use-lm-rescoring=1 --num-path=-1 --max-duration=10 --output-beam-size=8

2021-04-11 10:37:58,913 INFO [common.py:356] [test-clean] %WER 5.72% [3008 / 52576, 562 ins, 164 del, 2282 sub ]
2021-04-11 10:46:03,670 INFO [common.py:356] [test-other] %WER 15.71% [8224 / 52343, 1331 ins, 562 del, 6331 sub ]
```

#### LM rescoring with n-best list

```
$ ./mmi_att_transformer_decode.py --use-lm-rescoring=1 --num-path=100 --max-duration=500 --output-beam-size=20

2021-04-11 15:17:07,792 INFO [common.py:356] [test-clean] %WER 6.31% [3316 / 52576, 746 ins, 160 del, 2410 sub ]
2021-04-11 15:19:48,583 INFO [common.py:356] [test-other] %WER 16.93% [8863 / 52343, 1649 ins, 514 del, 6700 sub ]
```

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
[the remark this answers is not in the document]
csukuangfj: Fixed by removal.

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -206,6 +224,12 @@ def main():
     max_duration = args.max_duration
     avg = args.avg
     att_rate = args.att_rate
+    use_lm_rescoring = args.use_lm_rescoring
+    num_paths = args.num_paths
+    if num_paths < 0:
+        num_paths = None
+    else:
+        assert num_paths > 1
```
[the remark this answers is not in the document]
csukuangfj: A new bool variable `use_whole_lattice` is introduced and `num_paths` is not changed anymore.

danpovey:
Great!!
When we do RNNLM rescoring with n-best lists, we'll have to figure out an
algorithm that is more accurate.
Probably it will involve a couple passes of rescoring paths.
Dan

snowfall/decoding/lm_rescore.py
pzelasko: Sorry I missed this reply. My question was this: in Kaldi, HCLG arc scores actually consisted of two types of scores: the LM scores, and "graph" scores, which came from pronunciation probabilities in lexiconp. AFAIK rescoring in Kaldi always first re-computed the LM scores with the original G and subtracted them from the LM scores in the lattice, so that the "graph" scores were preserved, and only then added the rescoring LM scores on top of that.

I am not sure if our HLG supports multiple pronunciations and their probabilities (or do we even need to do it anymore), but it's just something that came to my mind when reviewing.

danpovey:
Mm.
Currently we are not using pron-probs, although we could in the future.
We can easily attach additional cost types and store them as
differently-named attributes, if we need to in the future.
The FST-reading code already supports reading arbitrary numbers of fields.

snowfall/decoding/lm_rescore.py
```diff
@@ -2,10 +2,52 @@
...
+
+        fsas = k2.index(b_fsas, indexes)
+        b_to_a = k2.index(b_to_a_map, indexes)
+        path_lats = k2.intersect_device(a_fsas,
+                                        fsas,
+                                        b_to_a_map=b_to_a,
+                                        sorted_match_a=sorted_match_a)
+        ans.append(path_lats)
+
+    return k2.append(ans)
```
pzelasko, line 48: just wondering - shouldn't that be named `cat/concat` to be more consistent with numpy/torch/pandas?

csukuangfj: Yes, I agree with you. I was also once confused by this name. Only after I read its implementation did I realize it's actually equivalent to cat.

The current name is to match its counterpart in C++. Don't know what's Dan's opinion about this. @danpovey

danpovey (approved):
I think this should be good to merge!

danpovey:
I am fine to change both of them to cat.

PR #149, opened 2021-04-07
Refactoring: data pipeline

pzelasko:
Guys, take a look and let me know if you think this makes sense. I thought about this a bit and I think there are four main "modules" in snowfall:
- data pipeline (cuts/augmentation/dataloaders)
- models (e.g. conformer and its settings)
- training/testing loop structure
- loss functions (e.g. CTC, MMI, others and how they interact with lexicon/LM graphs)

I think they are 80% orthogonal and the 20% that interleaves is probably the tricky part to structure this in the right way. I think each of them should be structured into some sort of module that registers arguments in an `argparse.ArgumentParser`, and we can use `jsonargparse` later to keep the configuration in YAML files (this seems to work fine in ESPnet, it's also a bit similar to how Kaldi did it).

This PR is dealing with the data pipeline. I'm trying to put LibriSpeech and Aishell into a similar structure -- both recipes actually share 95% of the cuts-related code and another one that I started preparing for BABEL languages also seems to head in this direction.

The design is loosely inspired by [PytorchLightning's DataModules](https://pytorch-lightning.readthedocs.io/en/latest/extensions/datamodules.html) -- if we ever decide that we want to use Lightning it's probably easy to adopt it fully. I suggest first looking at `DataModule`, then `AsrDataModule`, then `LibriSpeechAsrDataModule`/`AishellAsrDataModule`, and then at the mmi_attention training/decoding scripts.

WDYT?

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -290,7 +290,7 @@ def main():
     # load dataset
     librispeech = LibriSpeechAsrDataModule(args)
     test_sets = ['test-clean', 'test-other']
-    for test_set, test_dl in zip(test_sets, librispeech.train_dataloaders()):
+    for test_set, test_dl in zip(test_sets, librispeech.test_dataloaders()):
         logging.info(f'* DECODING: {test_set}')
 
         results = decode(dataloader=test_dl,
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -11,13 +11,10 @@ import numpy as np
 import os
 import torch
 from k2 import Fsa, SymbolTable
-from kaldialign import edit_distance
 from pathlib import Path
 from typing import List
 from typing import Union
 
-from lhotse import CutSet, load_manifest
-from lhotse.dataset import K2SpeechRecognitionDataset, SingleCutSampler
 from snowfall.common import average_checkpoint, store_transcripts
 from snowfall.common import find_first_disambig_symbol
 from snowfall.common import get_texts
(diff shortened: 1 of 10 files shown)


egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -294,22 +288,11 @@ def main():
     HLG.requires_grad_(False)
 
     # load dataset
-    feature_dir = Path('exp/data')
+    librispeech = LibriSpeechAsrDataModule(args)
     test_sets = ['test-clean', 'test-other']
-    for test_set in test_sets:
+    for test_set, test_dl in zip(test_sets, librispeech.train_dataloaders()):
```
csukuangfj, line 293: Should it be `librispeech.test_dataloaders()`?

pzelasko: Thanks, fixed. I'll re-run the whole thing to make sure it works alright.

csukuangfj:
Looks great to me.

snowfall/data/asr_datamodule.py
```diff
@@ -0,0 +1,211 @@
...
+            '--num-buckets',
+            type=int,
+            default=30,
+            help='The number of buckets for the BucketingSampler'
+                 '(you might want to increase it for larger datasets).')
+        group.add_argument(
+            '--concatenate-cuts',
+            type=str2bool,
+            default=True,
+            help='When enabled, utterances (cuts) will be concatenated '
```
csukuangfj, line 71: I am worried about the potential issue of the self-attention in the encoder part of the transformer/conformer model if different waves are concatenated.

It will definitely happen that frames in one wave can attend to frames from another wave
if concatenation is used since the encoder part does not use **masked** self-attention.

@danpovey @pzelasko  What do you think?

pzelasko: I think it’s worth checking the model performance by training with concatenation disabled. If the WER is better then you’re right, if the WER is worse then concatenation could act as a regularizer (the model has to learn to attend to the right frames). It could also have no effect, but I don’t think we know which is true as of now.

danpovey: My recollection is that we tested this and it was OK, but we should test again.  I believe the internal attention code may support a quadratic mask that tells you whether a particular frame can attend to a particular other frame;  in principle we could pass a mask of that type in, although it would change interfaces.

PR #151, opened 2021-04-10
Fix mmi bigram

danpovey:


danpovey:
Some fixes to the basic TDNN+LSTM model, post-refactoring.
It's not perfect but at least it runs (with the same WER as in the RESULTS.md).

egs/librispeech/asr/simple_v1/mmi_bigram_train.py
```diff
@@ -40,6 +40,34 @@
 
 den_scale = 1.0
 
+def encode_supervisions(supervisions: Dict[str, torch.Tensor],
```
pzelasko, line 43: I think this one could be imported from `snowfall.objectives.common` — unless it required some change for compatibility?

pzelasko: Oh, I see — the subsampling factor. Yeah, it’s behaviour seems to be model dependent, we might need to add sth like `.subsample_supervisions` as a model method in the future.

PR #162, opened 2021-04-16
Collect validation objf from all nodes in DDP training.

csukuangfj:


danpovey:
Great!  Merging.

danpovey:
FYI, this is hanging for me in the 2nd epoch.
I think this is expected though, because my version of lhotse was about a week old.  I am trying updating it.

csukuangfj:
> I think this is expected though

I believe so.  I think it hangs at the end of the first epoch in your case.
Update to the latest lhoste should help.

PR #164, opened 2021-04-16
Print default values in the help message.

### LINKED ISSUE #163: Print default args
Would it be hard to get argparse to print out default args when you do --help?

csukuangfj:
Closes #163


danpovey:
great, thanks!

Issue #165, opened 2021-04-16
Speeding up training


danpovey:
After having a look at nsys output, I think we are largely limited by latency of sequential operations in IntersectDevice, IntersectDense, GetForwardScores and GetBackwardScores (and of memory transfer when we invoke Array1::Back()).
I think there are two ways we can significantly reduce the time taken:

 - We can let the num and den FSAs be processed together by concatenating together the FsaVecs and calling IntersectDevice() just once, getting the tot_scores just once, and then post-processing ranges of the tot_scores.

 - IntersectDevice() is called when forming minibatches (intersecting with L and then with ctc_topo).  If we can somehow arrange to batch these up it would be more efficient.  It might not be super convenient code-wise, though.

PR #166, opened 2021-04-16
Invoke k2.intersect_dense and get_tot_scores only once in LFMMILoss

csukuangfj:
Related to #165 

Need to verify that it improves training speed.

### DIFF
--- a/snowfall/objectives/mmi.py
+++ b/snowfall/objectives/mmi.py
@@ -50,11 +50,7 @@ class LFMMILoss(nn.Module):
         # The following converts den_graphs.aux_labels
         # from torch.Tensor to k2.RaggedInt so that
         # we can use k2.append() later
-        ragged_shape = k2.ragged.regular_ragged_shape(
-            dim0=den_graphs.aux_labels.numel(), dim1=1).to(device)
-
-        den_graphs.aux_labels = k2.RaggedInt(ragged_shape,
-                                             den_graphs.aux_labels)
+        den_graphs.convert_attr_to_ragged_(name='aux_labels')
 
         num_den_graphs = k2.append([num_graphs, den_graphs])
 
--- a/snowfall/objectives/mmi.py
+++ b/snowfall/objectives/mmi.py
@@ -32,20 +32,65 @@ class LFMMILoss(nn.Module):
             texts: List,
             supervision_segments: torch.Tensor
     ) -> Tuple[torch.Tensor, int, int]:
-        num_graphs, den_graphs = self.graph_compiler.compile(texts, self.P)
+        num_graphs, den_graphs = self.graph_compiler.compile(
+            texts, self.P, replicate_den=False)
+
         dense_fsa_vec = k2.DenseFsaVec(nnet_output, supervision_segments)
 
-        num_lats = k2.intersect_dense(num_graphs, dense_fsa_vec, 10.0)
-        den_lats = k2.intersect_dense(den_graphs, dense_fsa_vec, 10.0)
+        device = num_graphs.device
 
-        num_tot_scores = num_lats.get_tot_scores(
-            log_semiring=True,
-            use_double_scores=True
-        )
-        den_tot_scores = den_lats.get_tot_scores(
-            log_semiring=True,
-            use_double_scores=True
-        )
+        num_fsas = num_graphs.shape[0]
+        assert dense_fsa_vec.dim0() == num_fsas
+
+        assert den_graphs.shape[0] == 1
+
+        # the aux_labels of num_graphs is k2.RaggedInt
+        # but it is torch.Tensor for den_graphs.
+        #
+        # The following converts den_graphs.aux_labels
+        # from torch.Tensor to k2.RaggedInt so that
+        # we can use k2.append() later
+        ragged_shape = k2.ragged.regular_ragged_shape(
+            dim0=den_graphs.aux_labels.numel(), dim1=1).to(device)
+
+        den_graphs.aux_labels = k2.RaggedInt(ragged_shape,
+                                             den_graphs.aux_labels)
+
+        num_den_graphs = k2.append([num_graphs, den_graphs])
+
+        # NOTE: The a_to_b_map in k2.intersect_dense must be sorted
+        # so the following reorders num_den_graphs.
+
+        # [0, 1, 2, ... ]
+        num_graphs_indexes = torch.arange(num_fsas, dtype=torch.int32)
+
+        # [num_fsas, num_fsas, num_fsas, ... ]
+        den_graphs_indexes = torch.tensor([num_fsas] * num_fsas,
+                                          dtype=torch.int32)
+
+        # [0, num_fsas, 1, num_fsas, 2, num_fsas, ... ]
+        num_den_graphs_indexes = torch.stack(
+            [num_graphs_indexes, den_graphs_indexes]).t().reshape(-1).to(device)
+
+        num_den_reordered_graphs = k2.index(num_den_graphs, num_den_graphs_indexes)
+
+        # [[0, 1, 2, ...]]
+        a_to_b_map = torch.arange(num_fsas, dtype=torch.int32).reshape(1, -1)
+
+        # [[0, 1, 2, ...]] -> [0, 0, 1, 1, 2, 2, ... ]
+        a_to_b_map = a_to_b_map.repeat(2, 1).t().reshape(-1).to(device)
+
+        num_den_lats = k2.intersect_dense(num_den_reordered_graphs,
+                                          dense_fsa_vec,
+                                          output_beam=10.0,
+                                          a_to_b_map=a_to_b_map)
+
+        num_den_tot_scores = num_den_lats.get_tot_scores(
+            log_semiring=True, use_double_scores=True)
+
+        num_tot_scores = num_den_tot_scores[::2]
+        den_tot_scores = num_den_tot_scores[1::2]
+
         tot_scores = num_tot_scores - self.den_scale * den_tot_scores
         tot_score, tot_frames, all_frames = get_tot_objf_and_num_frames(
             tot_scores,
--- a/snowfall/objectives/mmi.py
+++ b/snowfall/objectives/mmi.py
@@ -52,7 +52,7 @@ class LFMMILoss(nn.Module):
         # we can use k2.append() later
         den_graphs.convert_attr_to_ragged_(name='aux_labels')
 
-        num_den_graphs = k2.append([num_graphs, den_graphs])
+        num_den_graphs = k2.cat([num_graphs, den_graphs])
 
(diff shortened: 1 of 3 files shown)


csukuangfj:
It takes 53 minutes for the 0th epoch with one GPU, which is about 5 minutes less than before.

snowfall/objectives/mmi.py
```diff
@@ -32,20 +32,65 @@ def forward(
...
+        assert den_graphs.shape[0] == 1
+
+        # the aux_labels of num_graphs is k2.RaggedInt
+        # but it is torch.Tensor for den_graphs.
+        #
+        # The following converts den_graphs.aux_labels
+        # from torch.Tensor to k2.RaggedInt so that
+        # we can use k2.append() later
+        ragged_shape = k2.ragged.regular_ragged_shape(
+            dim0=den_graphs.aux_labels.numel(), dim1=1).to(device)
```
pzelasko, line 54: Just for me to make sure I understand, dim0 is the number of arcs (weights on arcs) and dim1 is the number of graphs (“batch size”)?

danpovey:
Great!

snowfall/objectives/mmi.py
csukuangfj: Let me give you a concrete example.

Suppose you have the following FSA,
```
0 1 1 0.2
1 2 -1 0.3
2
```
which has only two arcs and three states. Let's call the above Fsa `fsa`.
```
fsa.labels == torch.tensor([1, -1], dtype=torch.int32)
fsa.labels.shape[0] == fsa.arcs.num_elements()

fsa.scores == torch.tensor([0.2, 0.3], dtype=torch.float32)
fsa.scores.shape[0] == fsa.arcs.num_eleements()
```
Note that there are two arcs and each arc has a **SCALAR** attribute, `label`. So its `labels` is a 1-D tensor.

Each arc also has an attribute, `score`, which also happens to be a 1-D tensor.

We can assign an arbitrary attribute to the arcs of `fsa`. For instance, we can assign a **scalar**
to each arc. In this case, since there are 2 arcs and each arc has a scalar attribute, we use
a 1-D tensor to represent this attribute, i.e.,
```
fsa.attr1 = torch.tensor([10, 20]) 

# It doesn't have to be of dtype torch.int32.
# The ONLY constraint is that `attr_tensor.shape[0] == num_arcs`
```

Also, we can assign a 1-D **TENSOR** (with 10 elements) to each arc. In this case, the resulting attribute is a 2-D tensor, i.e.,

```
fsa.attr2 = torch.rand(2, 10)

# It has to satisfy fsa.attr2.shape[0] == num_arcs
# Otherwise, it will raise an exception.
```

**CAUTION**: If the attribute is a torch.Tensor, then it requires that each arc's attribute has the same number of elements.
In the above case,  the first arc's attr2 has 10 entries, so is the second arc.

You can assign an N-D tensor attribute to each arc, if you like. The above examples demostrate `N==0` and `N==1`.


**WHAT** if different arcs have a different number of elements for a given attribute? 
In this case, we cannot use torch.Tensor anymore.
Instead, we can use `k2.RaggedInt` (`k2.RaggedFloat` can also be used, though it is not implemented in k2).

For instance, we give a third attribute `attr3` to `fsa`. The first arc's attribute value is [10, 20, 30] and the second one is [100]. It can be represented in k2 as
```
fsa.attr3 = k2.RaggedInt('[ [10 20 30] [100] ]')
```

We have
```
fsa.attr3.dim0() == fsa.arcs.num_elements()
```

Sometimes it may happen that some arcs's attribute value is empty. The following is an example:

```
fsa.attr4 = k2.RaggedInt('[ [1 2 3 4 5] [] ]')

fsa.attr4.dim0() == fsa.arcs.num_elements()
```
The first arc's attribute value is `[1, 2, 3, 4, 5]`, while the second arc's attribute value is empty.

---

Back to the current pull-request.

Suppose `den_graphs.aux_labels` is `torch.tensor([1, -1])`, the code changes it to
```
den_graphs.aux_labels = k2.RaggedInt('[ [1] [-1] ]')
```

The return value of
```
ragged_shape = k2.ragged.regular_ragged_shape(
            dim0=den_graphs.aux_labels.numel(), dim1=1).to(device)
```

is a `k2.RaggedShape`.  `dim0` represents the outer dimension, which is the number of arcs in this case.
The resulting ragged shape looks like
```
[ [x] [x] ]
```

There are two `[]` inside, which is specified by `dim0`.

There is one `x` inside each `[]`, which is specified by `dim1`.

Hope the above comment helps.

---

We are planning to write tutorials for Ragged tensors in k2. Please just wait for a few days or weeks.

csukuangfj:
```
./mmi_att_transformer_train.py --world-size=2 --master-port=12355 --num-epochs=10
```
Tensorboard training log can be found at https://tensorboard.dev/experiment/GXs3ByIRQamnyesw1GmE2g/

It took 3 hours 45 minutes to train for 10 epochs with **2** GPUs.

WERs of this pull-request:
```
# (averaged using the last 5 epochs)
$ ./mmi_att_transformer_decode.py
2021-04-16 20:31:30,977 INFO [common.py:364] [test-clean] %WER 6.75% [3551 / 52576, 446 ins, 306 del, 2799 sub ]
2021-04-16 20:33:40,812 INFO [common.py:364] [test-other] %WER 18.25% [9555 / 52343, 1038 ins, 940 del, 7577 sub ]
```

csukuangfj:
Will try to run nsys to compare the training time with/without this pull-request.

snowfall/objectives/mmi.py
pzelasko: Thank you, that's a very clear description. It's impressive how much work you guys put into supporting ragged shapes in many possible ways.

csukuangfj:
Here are the nsys profiling results.

Steps to reproduce:

1. Uncomment the following lines:
https://github.com/k2-fsa/snowfall/blob/9648ce14f9fe147680c7b8f743bce4b1e1d159d4/egs/librispeech/asr/simple_v1/mmi_att_transformer_train.py#L257-L259

2. Use the following command to get `*.qdrep` profiling file:

```
(bash) $ nsys profile ./mmi_att_transformer_train.py --master-port=12355
```

- qdrep for this pullquest (new.qdrep):  https://drive.google.com/file/d/1UzIqsbKhPIq0vKjuAtNGqztS0cn7VQSy/view?usp=sharing
- qdrep for the master branch (old.qdrep): https://drive.google.com/file/d/1eNgnlXCL7M-qRF2zFCmX2a84HFkuJOC7/view?usp=sharing

---

Training log for the profiling commands

```
# this pullrequest
$ nsys profile ./mmi_att_transformer_train.py --master-port=12355
Collecting data...
.......
2021-04-19 10:03:00,576 INFO [mmi_att_transformer_train.py:237] (0/1) batch 0, epoch 0/10 global average objf: 1.686591 over 12273.0
frames (100.0% kept), current batch average objf: 1.686591 over 12273 frames (100.0% kept) avg time waiting for batch 31.238s
2021-04-19 10:03:00,590 INFO [distributed.py:607] (0/1) Reducer buckets have been rebuilt in this iteration.
2021-04-19 10:03:13,159 INFO [mmi_att_transformer_train.py:237] (0/1) batch 10, epoch 0/10 global average objf: 1.739954 over 134424.
0 frames (100.0% kept), current batch average objf: 1.715883 over 12243 frames (100.0% kept) avg time waiting for batch 3.130s
Exiting early to get profile info
Processing events...
```

```
# master branch
$ nsys profile ./mmi_att_transformer_train.py --master-port=12355
Collecting data...
......
2021-04-19 10:07:54,980 INFO [mmi_att_transformer_train.py:237] (0/1) batch 0, epoch 0/10 global average objf: 1.678893 over 12273.0
frames (100.0% kept), current batch average objf: 1.678893 over 12273 frames (100.0% kept) avg time waiting for batch 30.193s
2021-04-19 10:07:54,994 INFO [distributed.py:607] (0/1) Reducer buckets have been rebuilt in this iteration.
2021-04-19 10:08:07,741 INFO [mmi_att_transformer_train.py:237] (0/1) batch 10, epoch 0/10 global average objf: 1.739565 over 134424.0 frames (100.0% kept), current batch average objf: 1.711411 over 12243 frames (100.0% kept) avg time waiting for batch 3.025s
Exiting early to get profile info
Processing events...
```

You can see that both commands process the same number of frames, i.e., process the same data, so the profiling results
are comparable between the two runs.

---

The following two screenshots are the visualization results of the above two qdrep files.
You can see that the total kernel execution time is comparable between the two runs for `k2.intersect_dense`. But this pull-request invokes `k2.intersect_dense` only once, while the master branch invokes `k2.intersect_dense` twice.

## this pull-request
new

## master branch
old

csukuangfj:
Ready for review and merge.

snowfall/objectives/mmi.py
```diff
@@ -32,20 +32,65 @@ def forward(
...
+
+        den_graphs.aux_labels = k2.RaggedInt(ragged_shape,
```
danpovey, line 56: I think it would be more elegant to make this a function invoked like `den_graphs.convert_attr_to_ragged('aux_labels', remove_eps=True)`.
I imagine this will still have [-1] on the final-arcs, which is probably what we want
as the num graphs likely have this.

danpovey: ... Incidentally, the opposite to this function would, I guess, be convert_attr_to_linear(), which would have to call ExpandArcs().  However this would have to be a function in fsa_algo, as in: `x = k2.convert_attr_to_linear(x, 'aux_labels')`, as it might modify the FSA.  We can just keep this in mind as something we might want to do in the future.

csukuangfj: > I think it would be more elegant to make this a function invoked like `den_graphs.convert_attr_to_ragged('aux_labels', remove_eps=True)`.
> 
> I imagine this will still have [-1] on the final-arcs, which is probably what we want
> 
> as the num graphs likely have this.
> 
> 
> 
> 

ok, will move it to k2.

csukuangfj: > den_graphs.convert_attr_to_ragged('aux_labels', remove_eps=True).

I assume that `aux_labels` for this function is a 1-D tensor. How do we handle the case for 2-D or N-D tensors?
Why do we need the extra argument `remove_eps`?

danpovey: Currently we don't have to support such a thing for N-d tensors (eventually we may decide to allow ragged+regular arrays like that).
The remove_eps is because in general when we store labels as ragged arrays, we remove the zero elements.  
In this particular situation it doesn't matter because we won't actually be doing anything with these labels.

danpovey:
I guess we can merge this now..

csukuangfj:
Now it depends on https://github.com/k2-fsa/k2/pull/723

Issue #169, opened 2021-04-21
Conformer & convergence


danpovey:
Got this via email from @zhu-han ...
```
I got the first reasonable result:2021-04-21 02:28:00,831 INFO [common.py:365] [test-clean] %WER 13.39% [7041 / 52576, 887 ins, 965 del, 5189 sub ]2021-04-21 02:29:42,636 INFO [common.py:365] [test-other] %WER 35.57% [18619 / 52343, 1327 ins, 3534 del, 13758 sub ]
The training log is  in the attachment. And code of this version is in https://github.com/zhu-han/snowfall/commit/4d4a0c42c175571e396736c757ceb6698afc9b18 
The differences with the original version in the paper are:# this version vs original version1) 4× subsampling vs 8× subsampling; 2) kernel size 3 vs kernel size 5;
The model could not converge without these two changes. But the performance gap between this and conformer is still large. Do you have some advice on that?
```
For convergence problems: we are working on a way to make tuings converge much easier..  @csukuangfj was going to commit it.  It involves using a simpler model as an alignment model, adding its output with the model being trained with a scale less than one, near the start of training.

PR #172, opened 2021-04-22
Improve training speed by pre-computing `compose(ctc_topo, P, L)`

csukuangfj:
Seems to be working, but it needs more tests.

Will continue with it after fixing https://github.com/k2-fsa/snowfall/issues/169 

---

Relates to #165  and depends on https://github.com/k2-fsa/k2/pull/726

### DIFF
--- a/snowfall/training/mmi_graph.py
+++ b/snowfall/training/mmi_graph.py
@@ -261,7 +261,6 @@ class MmiTrainingGraphCompiler(object):
                                          b_to_a_map,
                                          sorted_match_a=True)
         num_graphs = k2.invert(num_graphs)
-        assert num_graphs.requires_grad == P.requires_grad
 
         # TODO(fangjun): k2.connect supports only CPU.
         # Add CUDA support.
--- a/snowfall/training/mmi_graph.py
+++ b/snowfall/training/mmi_graph.py
@@ -61,16 +61,15 @@ class MmiTrainingGraphCompiler(object):
           Out of vocabulary word.
         '''
         self.lexicon = lexicon
-        L_inv = self.lexicon.L_inv.to(device)
+        L = self.lexicon.L.to(device)
 
-        if L_inv.properties & k2.fsa_properties.ARC_SORTED != 0:
-            L_inv = k2.arc_sort(L_inv)
+        L = k2.arc_sort(L)
 
-        assert L_inv.requires_grad is False
+        assert L.requires_grad is False
 
         assert oov in self.lexicon.words
 
-        self.L_inv = L_inv
+        self.L = L
         self.oov_id = self.lexicon.words[oov]
         self.device = device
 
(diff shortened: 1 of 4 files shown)


pzelasko:
I wonder if it makes sense to retain a copy of the un-optimized version of the LFMMI loss, maybe sth like “SimpleLFMMI”, as a reference for people who just want to understand how it works.

snowfall/training/mmi_graph.py
```diff
@@ -110,49 +109,194 @@ def compile(self,
...
+
+        num_graphs = k2.intersect_device(self.HPL_inv_sorted,
+                                         linear_fsas_with_self_loops,
+                                         b_to_a_map,
+                                         sorted_match_a=True)
+        num_graphs = k2.invert(num_graphs)
+
+        # TODO(fangjun): k2.connect supports only CPU.
+        # Add CUDA support.
+        num_graphs = k2.connect(num_graphs.to('cpu')).to(P.device)
```
csukuangfj, line 267: @danpovey 

I think we probably need a CUDA version of k2.connect(),
though I have not profiled this pull-request. It is currently
slower than before. Not sure if it is the problem of TaskRedirect or is caused by this statement.

danpovey: is .connect() really necessary?

csukuangfj: > is .connect() really necessary?

If I don't invoke `k2.connect`, then the resulting `get_tot_scores` for the num_lats returns all `-inf`.
If `k2.connect` is used, then `get_tot_scores` returns no `-inf`s.

danpovey:
can you find out how the num-states changes?
It could be a state-sorting issue, although we should be detecting that
from the property flags.

danpovey:
.. also, while non-connected input could cause search errors, I wouldn't
expect the result to be *all* infinity, unless there was a problem like
it was not state-sorted.

csukuangfj:
> .. also, while non-connected input could cause search errors, I wouldn't
expect the result to be *all* infinity, unless there was a problem like
it was not state-sorted.

Thanks, will check that.

csukuangfj:
@danpovey 

> . also, while non-connected input could cause search errors, I wouldn't
expect the result to be *all* infinity, unless there was a problem like
it was not state-sorted.

I confirm that the tot_scores of `num_lats` are all `-inf` even if it is top sorted.
(The number of Fsas in the FsaVec is 8).

---

Some information about the FsaVec before and after calling `k2.connect`:

## Before
num_fsas: 8
num_states: 3309424
num_arcs: 3315379
properties: "Valid|Nonempty|MaybeAccessible"

## after
num_fsas: 8
num_states: 3228
num_arcs: 9183
properties: "Valid|Nonempty|TopSorted|MaybeAccessible|MaybeCoaccessible"


(NOTE: k2.arc_sort is called later for both cases)


---

The `GetCounts()` issue is fixed in https://github.com/k2-fsa/k2/pull/726/commits/d43f77ed167799c6335ee4616f4073031c996e91
and it is this `num_graphs` that results in errors in `GetCounts()`. After fixing it, `k2.top_sort` is applied to
this `num_graphs` to make it top sorted (but not connected).

snowfall/training/mmi_graph.py
danpovey: Something is not right here.
I think it may be a mistake to compose ctc_topo unless it's right at the end.  I believe ctc_topo expects to be composed with something that was epsilon-free and which then had epsilon self-loops added.  Because we are interpreting the epsilons on one side as "blank", which is in a sense a real symbol, things are a little subtle there.

danpovey: ... so I think it may be OK to compose L and P, and to compose that with the transcripts, but I'd leave ctc_topo until the end.

csukuangfj: > I think it may be a mistake to compose ctc_topo unless it's right at the end. I believe ctc_topo expects to be composed with something that was epsilon-free and which then had epsilon self-loops added.

I am using `intersect_device(ctc_topo_inv, P_with_self_loops).invert()` (equivalent to `compose(ctc_topo, P, treat_epslion_speciall=True)`.

There is no `0` (neither blank nor epsilons) in P, so I think it is correct.

csukuangfj:
Here are the profiling results of this pull-request.


## this pull-request
Screen Shot 2021-04-25 at 7 01 11 PM

## master branch
Screen Shot 2021-04-25 at 7 01 18 PM

---

Even though this pull-request requires one less call to k2.intersect_device, it is actually slower. I compared the size of the resulting num_graphs, listed in below. You can see that the resulting num_graphs of this pull-request is actually larger, so it takes more time.  

## this pull-request
- num_fsas: 42
- num_states: 19626
- num_arcs: 51152

## master branch
- num_fsas: 42
- num_states: 8051
- num_arcs: 11051

---

Closing.

PR #177, opened 2021-04-25
Fix att_rate.

csukuangfj:


danpovey:
I'm fine to merge this and keep that code working even though we're not really using it.

PR #181, opened 2021-04-29
Vgg frontend

danpovey:


danpovey:
This is an attempt to reproduce the VGG frontend mentioned in this Facebook paper https://arxiv.org/pdf/2005.09150v2.pdf, more details here https://arxiv.org/pdf/1910.09799.pdf (although it is not 100% explicit).
With 2 training jobs, the following is the WER I got:
```
2021-04-29 01:22:22,443 INFO [common.py:181] (0/1) Save checkpoint to exp-conformer-noam-mmi-att-musan-sa-vgg/epoch-9.pt: epoch=9, learning_rate=0.00044242867775131045, objf=0.15493929661339992, valid_objf=0.2033769116459806
2021-04-29 10:59:50,561 INFO [common.py:380] [test-clean] %WER 6.76% [3554 / 52576, 488 ins, 284 del, 2782 sub ]
2021-04-29 11:01:57,991 INFO [common.py:380] [test-other] %WER 17.50% [9159 / 52343, 1090 ins, 793 del, 7276 sub ]
```
I believe this is a little better than our current results (the RESULTS.md currently has 6.90%/17.89% as the best results)  but it would be good if someone can post their current no-lm-rescoring baseline with 2 jobs, if they have it.

Caution: if this PR is applied, we would no longer be able to decode with current models and would have to retrain.

snowfall/models/transformer.py
```diff
@@ -385,6 +385,73 @@ def forward(self, x: Tensor) -> Tensor:
...
+        layers = []
+        block_dims = [32,64]
+
+        # The decision to use padding=1 for the 1st convolution, then padding=0
+        # for the 2nd and for the max-pooling, and ceil_mode=True, was driven by
+        # a back-compatibility concern so that the number of frames at the
+        # output would be equal to:
+        #  (((T-1)//2)-1)//2.
+        # We can consider changing this by using padding=1 on the 2nd convolution,
+        # so the num-frames at the output would be T//4.
```
pzelasko, line 421: Would it make sense to add a method to each model that computes the subsampled shape from an original shape (or dim)? It’d be easier to support differing subsampling strategies.

zhu-han: I think another option is to keep our models subsampling exactly with the value `subsampling_factor`,  like `T // 4` here. This would make the subsampling easier. We only need to pass the `subsampling_factor` to the corresponding function.

In this way, we'll need one side padding in some cases, like `Conv2dSubsampling` in our Transformer. The defalut convolution layers in PyTorch don't support one side padding. But `torch.nn.ConstantPad2d` could do this. We'll have to handle convolutions carefully in our model. Don't know if there will be any other side effects.

danpovey: I think something even easier would be to simply modify our code so it's insensitive to small differences in padding.   E.g. when computing the length of supervision segments we could just use T//4, but then truncate all segments so they fit within the actual computed output.  If this is possible?

pzelasko: I'm not sure if I understood your idea correctly. If you mean truncating `num_frames` in supervision segments when you expect them to be longer than the nnet output, I think we'd have to be careful and truncate it from both sides if a segment begins in the middle of a cut (i.e. shift the `start_frame` too).

danpovey: I guess I mean we can get the num_frames of the cuts always by dividing by 4, but then just post-process them so they never go beyond the bounds of the actual nnet output, in case that is slightly less than the original T//4.  That is-- not have any specific requirements on how our nnet handles the downsampling, but just deal with it after the fact.

danpovey: I think the easiest way to resolve this would be to add an option with default:
   `allow_truncate=0`
to DenseFsaVec.__init__() in k2, so it would truncate the last few frames from a segment if necessary to
make it fit within the nnet output; we could set this to 2 or 3, for instance.  It seems to me that this
would be very easy to implement.  The only thing that would not be quite right is in mmi.py where we do:
```
        tot_score, tot_frames, all_frames = get_tot_objf_and_num_frames(
            tot_scores,
            supervision_segments[:, 2]
        )
```
supervision_segments[:, 2] may not have quite the right lengths.  But we could easily add a member function to the
DenseFsaVec object in Python, that would provide the correct length (e.g. we could save something in the constructor, to avoid an extra GPU->CPU transfer).

danpovey: .. does anyone have time to do this in k2?   (or let me know if there are any objections).

csukuangfj: ok, I am creating a PR in k2.

danpovey:
Thanks!

danpovey:
FYI, the results after 4-gram LM rescoring are:
```
2021-05-01 16:49:54,049 INFO [common.py:380] [test-other] %WER 15.44% [8081 / 52343, 1283 ins, 494 del, 6304 sub ]
2021-05-01 16:15:39,294 INFO [common.py:380] [test-clean] %WER 5.75% [3024 / 52576, 572 ins, 164 del, 2288 sub ]
```
@zhu-han would you mind running this on your cluster and updating RESULTS.md, and making your own PR based on this, with the results?  Sorry, I'm too busy to properly finish this right now.

zhu-han:
Sure, I'll do it.

danpovey:
Merged via #182

PR #189, opened 2021-05-08
Rework code that deals with conformer positional embedding for clarity and efficiency

danpovey:


danpovey:
I trained a model with this and verified that the decoding was identical with this vs. master code, so I'm confident enough that there isn't a bug.

PR #197, opened 2021-05-16
Concatenate valid cuts, reduce chance of OOM

danpovey:


danpovey:
This prevents a problem where the training would work OK but there would be OOM while getting the valid prob (even though the validation should use less memory since it doesn't have to do backprop).
Caution: the valid objf values are slightly worse (by 0.01 or 0.02) so you shouldn't compare them before/after this change.

pzelasko:
Another option is to set a different (lower) max duration for the dev set (and maybe bucketing to make the computation less wasteful), which would keep the same loss values.

danpovey:
I think I'd rather keep the train and valid process similar.
Will merge this; I'm assuming Fangjun has no objection..

danpovey:
That's interesting.
I'm skeptical that this would make a difference to the perormance though.

PR #106, opened 2021-02-22
WIP: Compute expected times per pathphone_idx.

csukuangfj:
Closes #96 

@danpovey  Do you have any idea how to test the code? And I am not sure how the return value is used.

I am using `mbr_lats` instead of `den_lats` as I find that `phone_seqs` from `mbr_lats` contains more zeros than that from `den_lats`.

I don't quite understand the normalization step from https://github.com/k2-fsa/snowfall/issues/96#issue-805274300
so it is not done in the code.

> Multiply the posteriors by the frame_idx and then sum over the columns of the sparse matrix of posteriors to get the total (frame_idx * occupation prob) for each seqphone_idx, and divide by the total occupation_prob for each seqphone_idx

PR #198, opened 2021-05-18
[WIP] Implement 2nd pass training using 1-best decoding results from the 1st pass network

csukuangfj:
It implements https://github.com/k2-fsa/snowfall/pull/106#issuecomment-803796177

> BTW, since it seems this is hard to get to work, if you feel like it you could work on a simpler idea.
In training time we'd take the best-path alignment and using some kind of RNN or masked attention
we'd predict the next label in that best-path. (We'd probably take in the output of the 1st network
as an input to that). The sequence length here is the same as the original sequence length.
In test time the way this would work at least initially, is we'd run this on n-best lists obtained from the
1st-pass decoding and use the scores to decide which of the n-best paths to keep. There are more accurate
decoding methods we could look into later.

---

The training objf is decreasing and seems to be converging. Will post the decoding results later.

### DIFF
--- a/snowfall/models/second_pass_model.py
+++ b/snowfall/models/second_pass_model.py
@@ -1,0 +1,240 @@
+#!/usr/bin/env python3
+
+# Copyright (c)  2021  Xiaomi Corporation (authors: Fangjun Kuang)
+#
+import logging
+
+import torch
+import torch.nn as nn
+from snowfall.models.transformer import PositionalEncoding
+from snowfall.models.transformer import TransformerEncoderLayer
+from snowfall.models.transformer import generate_square_subsequent_mask
+
+import k2
+
+
+def _compute_padding_mask(len_per_seq: torch.Tensor):
+    '''Sequences are of different lengths in number of frames
+    and they are padded to the longest length. This function
+    returns a mask to exclude the padded positions for attention.
+
+    The returned mask is called `key_padding_mask` in PyTorch's
+    implementation of multihead attention.
+
+    Args:
+      len_per_seq:
+        A 1-D tensor of dtype torch.int32 containing the number
+        of entries per seq before padding.
+    Returns:
+      Return a bool tensor of shape (len_per_seq.shape[0], max(len_per_seq)).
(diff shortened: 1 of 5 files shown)


snowfall/models/second_pass_model.py
```diff
@@ -0,0 +1,240 @@
...
+                          mask=attn_mask,
+                          src_key_padding_mask=key_padding_mask)
+
+        # x2 is (T, B, F)
+
+        x2 = x2.permute(1, 0, 2)
+
+        # now x2 is (B, T, F)
+
+        x_concat = torch.cat((padded_acoustics, x2), dim=-1)
```
csukuangfj, line 217: TODO(fangjun): Use cross attention here, like decoders
- query: x2
- key and value: padded_acoustics

snowfall/common2.py
```diff
@@ -0,0 +1,484 @@
+#!/usr/bin/env python3
```
csukuangfj, line 1: common2.py is the same as common.py, except that it has some code supporting
the second pass model. To avoid conflicts with the master,  a new file is used.

The same goes for the following xxx2.py files, e.g., lm_rescore2.py, mmi2.py.

snowfall/decoding/second_pass.py
```diff
@@ -0,0 +1,208 @@
...
+
+# This file implements the ideas proposed by Daniel Povey.
+#
+# See https://github.com/k2-fsa/snowfall/issues/232 for more details
+#
+
+import k2
+
+
+class Nbest(object):
```
csukuangfj, line 11: @danpovey 

This file implements the `Nbest` class proposed in
https://github.com/k2-fsa/snowfall/issues/232#issue-941386780

danpovey: That's great!  Yes it looks like what I had in mind.
I assume you would separate it from this PR though?  Or maybe even submit it to k2?  Since there's a lot going on here.

csukuangfj: Will move it to k2.

PR #201, opened 2021-05-24
n-best rescore with transformer lm

glynpu:
Wer results of this pr (by loaded models from espnet model zoo):
```
test-clean 2.43% 
test-other 5.79%
```

image

image

This pr implements following procedure with models from espnet model zoo:
image

Added benefits by loading espnet  trained conformer encoder model with equivalent snowfall model definition:
1. identify differences of conformer implementation between espnet and snowfall. As shown in snowfall/models/conformer.py, snowfall only scaling q; while espnet scale attn_outout_weights.
2. espnet conformer has an extra layer_norm after encoder

Also, the loaded espnet transformer lm could be used as a baseline for snowfall lm training tasks.

### DIFF
--- a/egs/librispeech/asr/simple_v1/espnet_utils/asr.py
+++ b/egs/librispeech/asr/simple_v1/espnet_utils/asr.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+
+# Copyright 2021 Xiaomi Corporation (Author: Guo Liyong)
+# Apache 2.0
+
+import argparse
+import logging
+from typing import Tuple
+
+import numpy as np
+import torch
+
+from espnet_utils.common import load_espnet_model_config
+from espnet_utils.common import rename_state_dict, combine_qkv
+from espnet_utils.frontened import Fbank
+from espnet_utils.frontened import GlobalMVN
+from espnet_utils.numericalizer import SpmNumericalizer
+from snowfall.models.conformer import Conformer
+
+_ESPNET_ENCODER_KEY_TO_SNOWFALL_KEY = [
+    ('frontend.logmel.melmat', 'frontend.melmat'),
+    ('encoder.embed.out.0.weight', 'encoder.embed.out.weight'),
+    ('encoder.embed.out.0.bias', 'encoder.embed.out.bias'),
+    (r'(encoder.encoders.)(\d+)(.self_attn.)linear_out([\s\S*])',
+     r'\1\2\3out_proj\4'),
+    (r'(encoder.encoders.)(\d+)', r'\1layers.\2'),
+    (r'(encoder.encoders.layers.)(\d+)(.feed_forward.)(w_1)',
+     r'\1\2.feed_forward.0'),
+    (r'(encoder.encoders.layers.)(\d+)(.feed_forward.)(w_2)',
+     r'\1\2.feed_forward.3'),
+    (r'(encoder.encoders.layers.)(\d+)(.feed_forward_macaron.)(w_1)',
+     r'\1\2.feed_forward_macaron.0'),
+    (r'(encoder.encoders.layers.)(\d+)(.feed_forward_macaron.)(w_2)',
+     r'\1\2.feed_forward_macaron.3'),
+    (r'(encoder.embed.)([\s\S*])', r'encoder.encoder_embed.\2'),
+    (r'(encoder.encoders.)([\s\S*])', r'encoder.encoder.\2'),
+    (r'(ctc.ctc_lo.)([\s\S*])', r'encoder.encoder_output_layer.1.\2'),
+]
+
+
+class ESPnetASRModel(torch.nn.Module):
+
+    def __init__(
+        self,
+        frontend: None,
+        normalize: None,
+        encoder: None,
+    ):
+
+        super().__init__()
+        self.frontend = frontend
+        self.normalize = normalize
+        self.encoder = encoder
+
+    def encode(
+            self, speech: torch.Tensor,
+            speech_lengths: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+
+        feats, feats_lengths = self.frontend(speech, speech_lengths)
+
+        feats, feats_lengths = self.normalize(feats, feats_lengths)
+
+        feats = feats.permute(0, 2, 1)
+
+        nnet_output, _, _ = self.encoder(feats)
+        nnet_output = nnet_output.permute(2, 0, 1)
+        return nnet_output
+
+    @classmethod
+    def build_model(cls, asr_train_config, asr_model_file, device):
+        args = load_espnet_model_config(asr_train_config)
+        # {'fs': '16k', 'hop_length': 256, 'n_fft': 512}
+        frontend = Fbank(**args.frontend_conf)
+        normalize = GlobalMVN(**args.normalize_conf)
+        encoder = Conformer(num_features=80,
+                            num_classes=len(args.token_list),
+                            subsampling_factor=4,
+                            d_model=512,
+                            nhead=8,
+                            dim_feedforward=2048,
+                            num_encoder_layers=12,
+                            cnn_module_kernel=31,
+                            num_decoder_layers=0,
+                            is_espnet_structure=True)
+
+        model = ESPnetASRModel(
+            frontend=frontend,
+            normalize=normalize,
+            encoder=encoder,
+        )
+
+        state_dict = torch.load(asr_model_file, map_location=device)
+
+        state_dict = {
+            k: v for k, v in state_dict.items() if not k.startswith('decoder')
+        }
+
+        combine_qkv(state_dict, num_encoder_layers=11)
+        rename_state_dict(rename_patterns=_ESPNET_ENCODER_KEY_TO_SNOWFALL_KEY,
+                          state_dict=state_dict)
+
+        model.load_state_dict(state_dict, strict=False)
+        model = model.to(torch.device(device))
+
+        numericalizer = SpmNumericalizer(tokenizer_type='spm',
+                                         tokenizer_file=args.bpemodel,
+                                         token_list=args.token_list,
+                                         unk_symbol='<unk>')
+        return model, numericalizer
--- a/egs/librispeech/asr/simple_v1/espnet_utils/frontened.py
+++ b/egs/librispeech/asr/simple_v1/espnet_utils/frontened.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python3
+
+# Copyright 2021 Xiaomi Corporation (Author: Guo Liyong)
+# Apache 2.0
+
+import humanfriendly
+import librosa
+import numpy as np
+import torch
+
+from pathlib import Path
+from typeguard import check_argument_types
+from typing import Optional, Tuple, Union
+
+
+# Modified from:
+# https://github.com/espnet/espnet/blob/08feae5bb93fa8f6dcba66760c8617a4b5e39d70/espnet/nets/pytorch_backend/frontends/feature_transform.py#L135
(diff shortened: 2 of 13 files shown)


danpovey:
Great!!
I assume the modeling units are BPE pieces?  I think a good step towards resolving the difference would be to train 
(i) a CTC model
(ii) a LF-MMI model
using those same BPE pieces.

glynpu:
> Great!!
> I assume the modeling units are BPE pieces? I think a good step towards resolving the difference would be to train
> (i) a CTC model
> (ii) a LF-MMI model
> using those same BPE pieces.

Yes, the modeling units are 5000 tokens including <blank>. 
I will do the suggested experiments.

danpovey:
Thanks!
You may run into memory problems.  Fangjun recently committed a code change
that can be used to work around something related to that, though.
We need to make sure our recipes can run for those kinds of sizes anyway.

snowfall/decoding/lm_rescore.py
```diff
@@ -187,12 +192,16 @@ def rescore_with_n_best_list(lats: k2.Fsa, G: k2.Fsa,
...
-                                     sorted_match_a=True)
-    lm_path_lats = k2.top_sort(k2.connect(lm_path_lats.to('cpu'))).to(device)
-    lm_scores = lm_path_lats.get_tot_scores(True, True)
+    if G is not None:
+        lm_path_lats = _intersect_device(G,
+                                         word_fsas_with_epsilon_loops,
+                                         b_to_a_map=b_to_a_map,
+                                         sorted_match_a=True)
+        lm_path_lats = k2.top_sort(k2.connect(lm_path_lats.to('cpu'))).to(device)
+        lm_scores = lm_path_lats.get_tot_scores(True, True)
```
danpovey, line 201: The 2nd arg to get_tot_scores() here should be false, because ARPA-type language models are constructed in such a way that the backoff prob is included in the direct arc.  I.e. we would be double-counting if we were to sum the probabilities of the non-backoff and backoff arcs.

egs/librispeech/asr/simple_v1/espnet_utils/frontened.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # feat: (B, T, D)
+        if norm_means:
+            if x.requires_grad:
+                x = x - self.mean
+            else:
+                x -= self.mean
+
+        if norm_vars:
+            x /= self.std
```
csukuangfj, line 88: `norm_means` uses a guard `requires_grad` to choose whether to perform an in-place update. Is there a reason not to do the same here?

The original implementation
https://github.com/espnet/espnet/blob/08feae5bb93fa8f6dcba66760c8617a4b5e39d70/espnet/nets/pytorch_backend/frontends/feature_transform.py#L135
uses `self.scale` to do a multiplication, which is more efficient than dividing by `self.std`.

egs/librispeech/asr/simple_v1/espnet_utils/asr.py
```diff
@@ -0,0 +1,109 @@
...
+
+        super().__init__()
+        self.frontend = frontend
+        self.normalize = normalize
+        self.encoder = encoder
+
+    def encode(
+            self, speech: torch.Tensor,
+            speech_lengths: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+
```
csukuangfj, line 58: Would you mind adding doc describing the shape of various tensors?

egs/librispeech/asr/simple_v1/espnet_utils/asr.py
```diff
@@ -0,0 +1,109 @@
...
+        feats, feats_lengths = self.normalize(feats, feats_lengths)
+
+        feats = feats.permute(0, 2, 1)
+
+        nnet_output, _, _ = self.encoder(feats)
+        nnet_output = nnet_output.permute(2, 0, 1)
+        return nnet_output
+
+    @classmethod
+    def build_model(cls, asr_train_config, asr_model_file, device):
```
csukuangfj, line 70: `cls` is never used.
I would suggest changing `@classmethod` to `@staticmethod` and removing `cls`.

egs/librispeech/asr/simple_v1/espnet_utils/load_lm_model.py
```diff
@@ -0,0 +1,58 @@
...
+    """This method is used to load LM model downloaded from espnet model zoo.
+
+    Args:
+        config_file: The yaml file saved when training.
+        model_file: The model file saved when training.
+
+    """
+    model = TransformerLM(**config)
+
+    assert model_file is not None, f"model file doesn't exist"
```
csukuangfj, line 39: `f"{model_file} doesn't exist"`

egs/librispeech/asr/simple_v1/espnet_utils/load_lm_model.py
```diff
@@ -0,0 +1,58 @@
...
+    return model
+
+
+def build_lm_model_from_file(config=None,
+                             model_file=None,
+                             model_type='espnet'):
+    if model_type == 'espnet':
+        return load_espnet_model(config, model_file)
+    elif model_type == 'snowfall':
+        raise NotImplementedError(f'Snowfall model to be suppported')
```
csukuangfj, line 56: No need to use `f-string` here.

egs/librispeech/asr/simple_v1/espnet_utils/nnlm_evaluator.py
```diff
@@ -0,0 +1,113 @@
...
+from espnet_utils.text_dataset import DatasetOption, TextFileDataIterator, AuxlabelDataIterator, AbsLMDataIterator
+from espnet_utils.load_lm_model import build_lm_model_from_file
+from espnet_utils.numericalizer import get_numericalizer
+from snowfall.models.lm_transformer import TransformerLM
+
+# TODO(Liyong Guo): types may need to be supported ['text', 'token', 'token_id']
+_TYPES_SUPPORTED = ['text_file', 'auxlabel']
+
+
+def _validate_input_type(input_type: Optional[str] = None):
```
csukuangfj, line 22: `input_type` cannot be `Optional` according to your code.

egs/librispeech/asr/simple_v1/espnet_utils/nnlm_evaluator.py
```diff
@@ -0,0 +1,113 @@
...
+    def nll(self, text_source):
+        nlls = []
+        total_nll = 0.0
+        total_ntokens = 0
+        total_nwords = 0
+        for xs_pad, target_pad, word_lens, token_lens in self.dataset(
+                text_source):
+            xs_pad = xs_pad.to(self.device)
+            target_pad = target_pad.to(self.device)
+            nll = self.lm.nll(xs_pad, target_pad, token_lens)
```
csukuangfj, line 59: This `nll` is the same as the current method name `nll`. Could you choose a better name for it?

egs/librispeech/asr/simple_v1/espnet_utils/numericalizer.py
```diff
@@ -0,0 +1,98 @@
...
+from dataclasses import dataclass
+from typing import Iterable, List, Optional, Union
+from pathlib import Path
+
+import numpy as np
+
+from torchtext.data.functional import load_sp_model
+
+
+class NumericalizerMixin(ABC):
```
csukuangfj, line 16: Why does this class inherit from `ABC`?

egs/librispeech/asr/simple_v1/espnet_utils/numericalizer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _assign_special_symbols(self):
```
csukuangfj, line 18: Is there a reason that this function name starts with an underline?

egs/librispeech/asr/simple_v1/espnet_utils/numericalizer.py
```diff
@@ -0,0 +1,98 @@
...
+        # <sos> and <eos> share same index for model download from espnet model zoo
+        assert '<sos/eos>' in self.token2idx \
```
csukuangfj, line 20: At first sight, this class contains no data field at all; so the meaning of `self.token2idx`
is quite confusing here. Could you add more documentation here?

egs/librispeech/asr/simple_v1/espnet_utils/numericalizer.py
```diff
@@ -0,0 +1,98 @@
...
+        self.sos_idx = self.token2idx[
+            '<sos/eos>'] if '<sos/eos>' in self.token2idx else self.token2idx[
+                '<sos>']
+        self.eos_idx = self.token2idx[
+            '<sos/eos>'] if '<sos/eos>' in self.token2idx else self.token2idx[
+                '<eos>']
+        self.unk_idx = self.token2idx['<unk>']
+
+
+@dataclass
```
csukuangfj, line 32: Do we really need to use `dataclass` here?

Also, could you remove the class `NumericalizerMixin`?
The extra level of inheritance makes the code hard to read.

egs/librispeech/asr/simple_v1/nnlm_nbest_rescore.sh
```diff
@@ -0,0 +1,48 @@
...
+      exit 1
+    fi
+  done
+fi
+if [ $stage -le 1 ]; then
+  # Download espnet pretrained models
+  # The original link of these models is:
+  # https://zenodo.org/record/4604066#.YKtNrqgzZPY
+  # which is accessible by espnet utils
+  # The are ported to following link for users who don't have espnet dependencies.
```
csukuangfj, line 27: Nit: The -> They

egs/librispeech/asr/simple_v1/nnlm_nbest_rescore.sh
```diff
@@ -0,0 +1,48 @@
...
+  if [ ! -d snowfall_model_zoo ]; then
+    echo "About to download pretrained models."
+    git clone https://huggingface.co/GuoLiyong/snowfall_model_zoo
```
csukuangfj, line 30: I would suggest using `git clone --depth 1`. It improves the clone speed.

egs/librispeech/asr/simple_v1/tokenizer_ctc_att_transformer_decode.py
```diff
@@ -0,0 +1,211 @@
...
+        speech = speech.to(torch.device(device))
+        lengths = lengths.to(torch.device(device))
+
+        nnet_output = model.encode(speech=speech, speech_lengths=lengths)
+        nnet_output = nnet_output.detach()
+
+        blank_bias = -1.0
+        nnet_output[:, :, 0] += blank_bias
+
+        supervision_segments = torch.tensor([[0, 0, nnet_output.shape[1]]],
```
csukuangfj, line 63: Is the batch size always 1? A larger batch size can improve decoding speed.

egs/librispeech/asr/simple_v1/tokenizer_ctc_att_transformer_decode.py
```diff
@@ -0,0 +1,211 @@
...
+                [-1, 0, numericalizer.sos_idx, numericalizer.eos_idx],
+                best_paths.aux_labels.cpu().numpy()))
+
+        token = numericalizer.ids2tokens(token_int)
+
+        text = numericalizer.tokens2text(token)
+
+        ref = batch['supervisions']['text']
+        for i in range(len(ref)):
+            hyp_words = text.split(' ')
```
csukuangfj, line 93: What's the format of `text`?
Does `text` depend on `i`? If not, you can `split` it outside of the for loop.

csukuangfj (changes requested):
Please add more documentation to your code.

PR #207, opened 2021-06-05
Begin to add small tools for snowfall

csukuangfj:


csukuangfj:
From https://github.com/lhotse-speech/lhotse/pull/304#issuecomment-839408455

> And we might want some way to represent those data alignments. E.g. one simple representation might just be the label sequences, indexed somehow by utterance-id. There are several different label-sequences that might be relevant here, depending on the type of system: the ilabel from the model, the phone-label "without repetitions", which we could store as a separate attribute on the graphs by using the "inner_labels=xxx" arg to compose in the appropriate stage of graph creation; and the olabel which is the word label. One possibility is to just store these as a dict indexed by utterance-id and then by 'ilabel', 'olabel' and 'phone_label' (for phone labels without repetitions) or something like that, and store it as a .pt file with torch.save().

--- 

From https://github.com/lhotse-speech/lhotse/pull/304#discussion_r645978753

> Btw @csukuangfj we should make sure that Lhotse alignments and Snowfall alignments interact well together. If there are any additions/changes that are helpful on Lhotse side, please let us know.

---

I am adding a new class `Alignment` to implement the ideas from Dan. See
https://github.com/k2-fsa/snowfall/blob/692d5c426e232e9391fd2580011ebfc6b1e0c035/snowfall/tools/ali.py#L11-L19

If sampling rate is available and if the alignment is frame-wise, I think it's very easy to convert it to CTM, like the one supported by Lhotse.

csukuangfj:
The usage of the commanline tools is like lhotse.

1. Install `snowfall`
```
cd snowfall
python3 setup.py install
```

2. Invoke it
```
$ snowfall --help
Usage: snowfall [OPTIONS] COMMAND [ARGS]...

  Entry point to the collection of utilities in snowall.

Options:
  --help  Show this message and exit.

Commands:
  ali  Alignment tools in snowfall
```

```
$ snowfall ali --help
Usage: snowfall ali [OPTIONS] COMMAND [ARGS]...

  Alignment tools in snowfall

Options:
  --help  Show this message and exit.

Commands:
  edit-distance  Compute edit distance between two alignments.
```

```
$ snowfall ali edit-distance --help
Usage: snowfall ali edit-distance [OPTIONS]

  Compute edit distance between two alignments.

  The reference/hypothesis alignment file contains a python object Dict[str,
  Alignment] and it can be loaded using `torch.load`. The dict is indexed by
  utterance ID.

  The symbol table, if provided, has the following format for each line:

      symbol integer_id

  It can be loaded by `k2.SymbolTable.from_file()`.

Options:
  -r, --ref FILE           The file containing reference alignments
                           [required]
  -h, --hyp FILE           The file containing hypothesis alignments
                           [required]
  -t, --type TEXT          The type of the alignment to use for computing the
                           edit distance  [required]
  -o, --output-file FILE   Output file
  -s, --symbol-table FILE  The symbol table for the given type of alignment
  --help                   Show this message and exit.
```

csukuangfj:
Ready for review.

snowfall/tools/ali.py
```diff
@@ -0,0 +1,88 @@
...
+        The type of alignment to use.
+      output_file:
+        The filename of the output file.
+    Returns:
+      Return None.
+    '''
+    pairs = []  # each element contains a pair (ref_transcript, hyp_transcript)
+
+    # We assume that the same utterance exists both in ref_ali and in hyp_ali.
+    # An exception is thrown if the assumption is violated.
```
danpovey, line 81: Sometimes decoding fails for a single utterance... might be better to just print a warning if some are missing from one or the other?

test/fixtures/gen_ali.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
+
```
danpovey, line 2: Does some kind of test program go with this that you haven't checked in?

csukuangfj: The following is how I run the test:

```bash
$ cd snowfall/test/fixtures
$ ./gen_ali.py
# It generates: hyp.pt, ref.pt, sym.txt

$ snowfall ali edit-distance  -r ./ref.pt -h ./hyp.pt -t phone -o a.txt -s ./sym.txt
# It generates the output file: a.txt
# I checked manually that `a.txt` contains the expected result.
```

---

The files in `snowfall/bin` contain simple wrappers around the existing code. So I don't write unit tests
for command-line tools. Also, snowfall currently contains only a few unit tests.

snowfall/tools/ali.py
csukuangfj: The code is updated to print a warning if an utterance exists in the reference but has no corresponding hypothesis.

test/fixtures/gen_ali.py
danpovey: Ah, cool.  Perhaps can note that down in a file checked in somewhere?  Like a README.md somehwere in the cli directory, if there is one?

snowfall/tools/ali.py
```diff
@@ -0,0 +1,100 @@
...
+    '''
+    pairs = []  # each element contains a pair (ref_transcript, hyp_transcript)
+
+    utts_in_ref = set(ref_ali.keys())
+    utts_in_hyp = set(hyp_ali.keys())
+
+    utts_in_both = utts_in_ref & utts_in_hyp
+
+    utts_in_ref_only = utts_in_ref - utts_in_hyp
+    if utts_in_ref_only:
```
danpovey, line 88: can we make this symmetric with ref and hyp, so print warnings in both cases?

test/fixtures/gen_ali.py
danpovey: .. and is it possible to run that 'snowfall' command without using setup.py to install?

csukuangfj: > .. and is it possible to run that 'snowfall' command without using setup.py to install?

Yes, it is possible.

There are two ways.
(1)
```
cd snowfall
python3 ./snowfall/bin/snowfall
```


(2)
```
chmod +x /path/to/snowfall/snowfall/bin/snowfall
export PATH=/path/to/snowfall/snowfall/bin:$PATH
# then you can run
snowfall --help
```

The following is my setting (without installing k2, snowfall, and lhotse)

```bash
export PYTHONPATH=$HOME/open-source/k2/build_release/lib:$PYTHONPATH
export PYTHONPATH=$HOME/open-source/k2/k2/python:$PYTHONPATH
export PYTHONPATH=$HOME/open-source/snowfall:$PYTHONPATH
export PYTHONPATH=$HOME/open-source/lhotse:$PYTHONPATH
export PATH=$HOME/open-source/snowfall/snowfall/bin:$PATH
```

snowfall/tools/ali.py
csukuangfj: Fixed.

test/fixtures/gen_ali.py
csukuangfj: I've added a file `snowfall/bin/README.rst`. Please see
https://github.com/k2-fsa/snowfall/blob/fcf0c53c555b89f8fff36d085f9a121c228b90d5/snowfall/bin/README.rst

danpovey:
LGTM!  Thanks for the very fast work!

PR #209, opened 2021-06-08
Use a modified ctc_topo.

csukuangfj:
This implements the topo mentioned in
https://github.com/k2-fsa/k2/issues/746#issuecomment-856421616

Some examples (assuming there are three phones: a, b, c)

Screen Shot 2021-06-08 at 4 16 26 PM

Screen Shot 2021-06-08 at 3 58 19 PM

Screen Shot 2021-06-08 at 3 58 25 PM

Screen Shot 2021-06-08 at 3 58 30 PM

Screen Shot 2021-06-08 at 3 58 35 PM

CAUTION: There are no mandatory blanks between the two consecutive symbols `aa` in `aabc`.

### DIFF
--- a/egs/librispeech/asr/simple_v1/run.sh
+++ b/egs/librispeech/asr/simple_v1/run.sh
@@ -31,23 +31,35 @@ fi
 
 if [ $stage -le 4 ]; then
   # Build G
-  python3 -m kaldilm \
-    --read-symbol-table="data/lang_nosp/words.txt" \
-    --disambig-symbol='#0' \
-    --max-order=1 \
-    data/local/lm/lm_tgmed.arpa >data/lang_nosp/G_uni.fst.txt
+  if [ ! -f data/lang_nosp/G_uni.fst.txt ]; then
+    python3 -m kaldilm \
+      --read-symbol-table="data/lang_nosp/words.txt" \
+      --disambig-symbol='#0' \
+      --max-order=1 \
+      data/local/lm/lm_tgmed.arpa >data/lang_nosp/G_uni.fst.txt
+  else
+    echo "Skip generating data/lang_nosp/G_uni.fst.txt"
+  fi
 
-  python3 -m kaldilm \
-    --read-symbol-table="data/lang_nosp/words.txt" \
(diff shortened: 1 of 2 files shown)


csukuangfj:
The following shows the same example using the existing ctc_topo:

Screen Shot 2021-06-08 at 4 05 22 PM
Screen Shot 2021-06-08 at 4 05 28 PM
Screen Shot 2021-06-08 at 4 05 33 PM
Screen Shot 2021-06-08 at 4 05 38 PM
Screen Shot 2021-06-08 at 4 05 44 PM

danpovey:
Wow that was fast!
Merge when you think it makes sense, looks good to me!
(BTW at some point we should change the args of those function to be just an integer saying the number of phones; the list input is not good because we require the list to be contiguous.)

pzelasko:
Maybe I’m missing sth but if we allow no blank between repeated phones, then isn’t the blank redundant? Can we simply use a 1 state pure self loop phone topo (+ final state) instead with the same result?

danpovey:
I think empirically the shared blank helps.  If the nnet doesn't want to
use it, it can just make it very improbable.
(in LF-MMI).

xiaohui-zhang:
Nice to see this non-standard topo implemented finally (as we discussed before @danpovey , Fig 1a in http://oa.ee.tsinghua.edu.cn/~ouzhijian/pdf/ctc-crf.pdf). This doesn't matter much during training (because even with the standard topo, we can use rule-based numerator FST construction rather than composing topo FST with tokenized transcripts, to lower the computation cost of numerator construction). But this significantly improves decoding speed (much smaller HLG). The WER degradation is minimal. The main issue is that words like "met" and "meet" will be more confusable during both training and decoding. @pzelasko Yeah having a shared blank is important regarding training performance, especially when we use specAug (we can achieve similar effects by allowing skippable silence phones within a word, which is hacky for HMM). But silence/HMM still has its advantage when we need a model to produce accurate alignments/decoding time-stamps.

pzelasko:
interesting, thanks for the explanation.

PR #210, opened 2021-06-10
WIP: add compute-post.

csukuangfj:
Usage:
```bash
$ snowfall net compute-post -m /ceph-fj/model-jit.pt -f exp/data/cuts_test-clean.json.gz -o exp
```

I find that there is one issue with the Torch Scripted module: We have to know the signature of the `forward` function
of the model, also have to know its subsampling factor.

----

Working on `compute-ali` and will submit them together.

### DIFF
--- a/snowfall/bin/modes/__init__.py
+++ b/snowfall/bin/modes/__init__.py
@@ -1,2 +1,3 @@
 from .ali import ali
 from .cli_base import cli
+from .net import net
--- a/snowfall/bin/modes/net.py
+++ b/snowfall/bin/modes/net.py
@@ -1,0 +1,134 @@
+# Copyright (c)  2021  Xiaomi Corp.       (author: Fangjun Kuang)
+
+from pathlib import Path
+
+import click
+import k2
+import lhotse
+import torch
+
+from .cli_base import cli
+
+
+@cli.group()
+def net():
+    '''
+    Neural network tools in snowfall.
+    '''
+    pass
+
+
+@net.command()
+@click.option('-m',
+              '--model',
+              type=click.Path(exists=True, dir_okay=False),
+              required=True,
+              help='Path to Torch Scripted module')
+@click.option('-f',
+              '--feats',
+              type=click.Path(exists=True, dir_okay=False),
+              required=True,
+              help='Path to Featureset manifest')
+@click.option('-l',
+              '--max-duration',
+              default=200,
+              type=int,
+              required=True,
+              help='max duration in seconds in a batch')
+@click.option('-i',
+              '--device-id',
+              default=0,
+              type=int,
+              required=True,
+              help='-1 to use CPU. Otherwise, it is the GPU device ID')
+@click.option('-o',
+              '--output-dir',
+              type=click.Path(dir_okay=True),
+              required=True,
+              help='Output directory')
+def compute_post(model: str, feats: str, max_duration: int, device_id: int,
+                 output_dir: str):
+    '''Compute posteriors given a model and a FeatureSet.
+    '''
+    scripted_model = torch.jit.load(model)
+    scripted_model.eval()
+
+    if device_id < 0:
+        print('Use CPU')
+        device = torch.device('cpu')
+    else:
+        print(f'Use GPU {device_id}')
+        device = torch.device('cuda', device_id)
+
+    scripted_model.to(device)
+
+    cuts = lhotse.load_manifest(feats)
+
+    dataset = lhotse.dataset.K2SpeechRecognitionDataset(cuts, return_cuts=True)
+
+    sampler = lhotse.dataset.SingleCutSampler(cuts, max_duration=max_duration)
+
+    dataloader = torch.utils.data.DataLoader(dataset,
+                                             batch_size=None,
+                                             sampler=sampler,
+                                             num_workers=1)
+
+    output_dir = Path(output_dir)
+    output_dir.mkdir(exist_ok=True)
+    storage_path = output_dir / 'posts'
+
+    posts_writer = lhotse.NumpyFilesWriter(storage_path=storage_path)
+    ans_cuts = []
+
+    for i, batch in enumerate(dataloader):
+        if i % 10 == 0:
+            print(f'Processing batch {i}')
+
+        feature = batch['inputs'].to(device)
+        feature_lens = batch['inputs_lens']
+
+        supervisions = batch['supervisions']
+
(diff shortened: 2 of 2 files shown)


csukuangfj:
I just created a pull-request in Lhotse https://github.com/lhotse-speech/lhotse/pull/319 to add
posteriors to the class `Cut`. The motivation is to reuse the serialization and dataset code from it.

---

Also, I find the alignment information contained in the supervision is too simple, see
https://github.com/lhotse-speech/lhotse/blob/ef7a037426f1b602a54f4d9ea43e711007e85719/lhotse/supervision.py#L24
```python
    symbol: str    
    start: Seconds    
    duration: Seconds
```

Can we move the alignment class from snowfall to lhotse?
https://github.com/k2-fsa/snowfall/blob/bce73304f40c321a6dad809058b12e559962c321/snowfall/tools/ali.py#L20-L28

csukuangfj:
The usage of `compute-ali`:
```bash
$ snowfall  ali compute-ali -l data/lang_nosp -p ./exp/cuts_post.json  --max-duration=500 -o exp
```

snowfall/tools/ali.py
```diff
@@ -196,3 +203,107 @@ def visualize(input: str,
...
+    lang_dir = Path(lang_dir)
+
+    symbol_table = k2.SymbolTable.from_file(lang_dir / 'words.txt')
+    phone_symbol_table = k2.SymbolTable.from_file(lang_dir / 'phones.txt')
+    phone_ids = get_phone_symbols(phone_symbol_table)
+
+    phone_ids_with_blank = [0] + phone_ids
+    ctc_topo = k2.arc_sort(build_ctc_topo(phone_ids_with_blank))
+
+    if not (lang_dir / 'HLG.pt').exists():
```
pzelasko, line 224: I think this could be refactored to a function and re-used across this script and decode scripts (and possibly others)

```python
def load_or_compile_HLG(lang_dir: Path) -> k2.Fsa: ...
```

snowfall/tools/ali.py
```diff
@@ -196,3 +203,107 @@ def visualize(input: str,
...
+                          labels_disambig_id_start=first_phone_disambig_id,
+                          aux_labels_disambig_id_start=first_word_disambig_id)
+        torch.save(HLG.as_dict(), lang_dir / 'HLG.pt')
+    else:
+        print('Loading pre-compiled HLG')
+        d = torch.load(lang_dir / 'HLG.pt')
+        HLG = k2.Fsa.from_dict(d)
+
+    HLG = HLG.to(device)
+    HLG.aux_labels = k2.ragged.remove_values_eq(HLG.aux_labels, 0)
```
pzelasko, line 246: what is this line doing? It looks like it's "sparsifying" the aux_labels (word ids) but how does HLG known which labels correspond to which aux_labels after that?

snowfall/tools/ali.py
```diff
@@ -196,3 +203,107 @@ def visualize(input: str,
...
+        indices = torch.argsort(supervision_segments[:, 2], descending=True)
+        supervision_segments = supervision_segments[indices]
+        texts = supervisions['text']  # TODO: remove it
+
+        sf = supervisions['cut'][0].posts.subsampling_factor
+        dense_fsa_vec = k2.DenseFsaVec(nnet_output,
+                                       supervision_segments,
+                                       allow_truncate=sf - 1)
+
+        lattices = k2.intersect_dense_pruned(HLG, dense_fsa_vec, 20.0,
```
pzelasko, line 284: the pruning related arguments here could be function parameters

snowfall/tools/net.py
```diff
@@ -0,0 +1,102 @@
...
+    dataloader = torch.utils.data.DataLoader(dataset,
+                                             batch_size=None,
+                                             sampler=sampler,
+                                             num_workers=1)
+
+    output_dir = Path(output_dir)
+    output_dir.mkdir(exist_ok=True)
+    storage_path = output_dir / 'posts'
+
+    posts_writer = lhotse.NumpyFilesWriter(storage_path=storage_path)
```
pzelasko, line 48: This is going to create a lot of files, not sure if NumpyHdf5Writer is preferable

snowfall/tools/ali.py
csukuangfj: Thanks! Will refactor it and add options to enable/disable LM rescoring.

pzelasko:
BTW I wonder if we should support piping these programs together, Kaldi-style. Click easily allows doing that with [file type arguments](https://click.palletsprojects.com/en/8.0.x/arguments/#file-arguments).

We could do that by writing/reading JSONL-serialized manifests in a streaming manner. Since most operations on `CutSet` refer to individual operations on `Cut`, this seems feasible without the need to re-write too much code. [There is a function in Lhotse](https://github.com/lhotse-speech/lhotse/blob/ef7a037426f1b602a54f4d9ea43e711007e85719/lhotse/serialization.py#L541) that tries to figure out the right manifest type from a dict, which can be used to parse individual lines (BTW @csukuangfj I just realized that you might need to extend that function to handle the posterior manifests in your Lhotse PR).

WDYT?

pzelasko:
... there is also some code for line-by-line [incremental JSONL writing in Lhotse](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/serialization.py#L189) that could be extended to support this.

danpovey:
This cool; I'm afraid I'm not following it in detail.
Just a reminder; this is more an "experimental direction" at this point.  We'll have to learn from experience whether these kinds of command line utilites are actually a useful thing.

pzelasko:
Fair enough. The idea is to allow sth like: 
```
snowfall net compute-post <some-inputs-args..> - | snowfall net compute-ali - <some-more-args..>
```
but I just realized that with the current way things are done in Lhotse, we would have store the actual arrays/tensors on disk and just pass the manifests around, which might not be optimal. Maybe it's not relevant for now and we can see how to do that in the future, if needed at all.

danpovey:
BTW, I tend to think being able to do something at all tends to be more important than that thing being efficient-- premature optimization being the root of all evil etc., although I did plenty of it in Kaldi.  I don't know  what the optimal solution is here, I am afraid I have not been following this PR closely enough.

pzelasko:
Agreed. But for the record, the full quote is actually: 

> "We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil. Yet we should not pass up our opportunities in that critical 3%."

PR #212, opened 2021-06-11
Add lm_scale to LM rescoring.

### LINKED ISSUE #211: Test results at multiple LM scales, picking best
Can we change the decoding script to test the results with multiple LM scales, say, 0.7 through 1.2 for now?

We already have the property 'lm_scores' on the lattices which dictates the LM part of the scores. (although 'scores' also contains the original acoustic scores plus the LM scores).  
We could clone the 'scores' array, and then repeatedly assign:
  `lats.scores = scores + lats.lm_scores * (lm_scale - 1.0)`
and then get the best paths again.
We will have to check that assigning to the 'scores' property in k2 properly invalidates the forward and backward
scores that are cached in the Fsa, otherwise all output will be the same.

For now let's just always output the WERs from each scale, and also a line that says "best WER" for each test set, so we can easily grep for it.

csukuangfj:
Closes #211 

@danpovey Please try this pull-request.

It requires a pull request I just created in k2 to clear the cache of an FAS if its scores are reassigned.

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -103,68 +103,6 @@ def decode(dataloader: torch.utils.data.DataLoader, model: AcousticModel,
     return results
 
 
-def print_transition_probabilities(P: k2.Fsa, phone_symbol_table: SymbolTable,
-                                   phone_ids: List[int], filename: str):
-    '''Print the transition probabilities of a phone LM.
-
-    Args:
-      P:
-        A bigram phone LM.
-      phone_symbol_table:
-        The phone symbol table.
-      phone_ids:
-        A list of phone ids
-      filename:
-        Filename to save the printed result.
-    '''
-    num_phones = len(phone_ids)
-    table = np.zeros((num_phones + 1, num_phones + 2))
-    table[:, 0] = 0
-    table[0, -1] = 0  # the start state has no arcs to the final state
-    assert P.arcs.dim0() == num_phones + 2
-    arcs = P.arcs.values()[:, :3]
-    probability = P.scores.exp().tolist()
-
-    assert arcs.shape[0] - num_phones == num_phones * (num_phones + 1)
-    for i, arc in enumerate(arcs.tolist()):
-        src_state, dest_state, label = arc[0], arc[1], arc[2]
-        prob = probability[i]
-        if label != -1:
-            assert label == dest_state
-        else:
-            assert dest_state == num_phones + 1
-        table[src_state][dest_state] = prob
-
-    try:
-        from prettytable import PrettyTable
-    except ImportError:
-        print('Please run `pip install prettytable`. Skip printing')
-        return
-
-    x = PrettyTable()
-
-    field_names = ['source']
-    field_names.append('sum')
-    for i in phone_ids:
-        field_names.append(phone_symbol_table[i])
-    field_names.append('final')
-
-    x.field_names = field_names
-
-    for row in range(num_phones + 1):
-        this_row = []
-        if row == 0:
-            this_row.append('start')
-        else:
-            this_row.append(phone_symbol_table[row])
-        this_row.append('{:.6f}'.format(table[row, 1:].sum()))
-        for col in range(1, num_phones + 2):
-            this_row.append('{:.6f}'.format(table[row, col]))
-        x.add_row(this_row)
-    with open(filename, 'w') as f:
-        f.write(str(x))
-
-
 def get_parser():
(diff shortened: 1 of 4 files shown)


egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -34,63 +41,95 @@
 from snowfall.training.mmi_graph import get_phone_symbols
 
 
+def decode_one_batch(batch: Dict[str, Any],
+                     model: AcousticModel,
+                     HLG: k2.Fsa,
+                     output_beam_size: float,
+                     num_paths: int,
+                     use_whole_lattice: bool,
+                     G: Optional[k2.Fsa] = None)->Dict[str, List[List[int]]]:
```
danpovey, line 50: should probably document this if we intend to check it in.

csukuangfj: I've added some documentation to it.

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -34,63 +41,138 @@
...
+        hyps = get_texts(best_paths, indices)
+        return {'no_rescore': hyps}
+
+    lm_scale_list = [0.7, 0.8, 0.9, 1.0, 1.1, 1.2]
+
+    ans = dict()
+    saved_scores = lattices.scores.clone()
+    for lm_scale in lm_scale_list:
+        lattices.scores = saved_scores.clone()
+        best_paths = decode_with_lm_rescoring(
```
danpovey, line 135: We could improve the speed here by first doing the composition, and then afterward doing the scaling.

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
```diff
@@ -34,63 +41,138 @@
...
+
+@torch.no_grad()
 def decode(dataloader: torch.utils.data.DataLoader, model: AcousticModel,
-           device: Union[str, torch.device], HLG: Fsa, symbols: SymbolTable,
+           HLG: Fsa, symbols: SymbolTable,
            num_paths: int, G: k2.Fsa, use_whole_lattice: bool, output_beam_size: float):
     tot_num_cuts = len(dataloader.dataset.cuts)
     num_cuts = 0
-    results = []  # a list of pair (ref_words, hyp_words)
+    results = defaultdict(list)
```
danpovey, line 154: can we have a comment about the structure of `results`?

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
csukuangfj: Thanks! I am fixing it.

danpovey:
Thanks!  you can merge when you think it's ready.  Then Desh can have another look at position-dependent phones with this in place.
BTW, in graph.py, when we do
`HLG.lm_scores = HLG.scores.clone()`
.. this might not be 100% correct because these also contain the silence probability at this point; we should perhaps be doing this at the G.fst level.  Of course, this depends on our precise rescoring method; but if we are subtracting the old lm_scores and composing with an n-gram FST, this would wrongly remove the silence probs.

aarora8:
Thank you so much, I will try to use it with position-dependent phones.

csukuangfj:
> we should perhaps be doing this at the G.fst level

Using G.lm_scores can indeed achieve a lower WER than that of HLG.lm_scores.

----

The following screenshots are the results applying `lm_scale` during rescoring.
(The model was trained using only a subset of the training data, i.e., only 100 hours of training data)

# Rescoring with whole lattice (HLG.lm_scores)

Screen Shot 2021-06-15 at 2 36 20 PM

It uses HLG.lm_scores. You can see that lm_scale helps to obtain a better WER 5.63 than the baseline 5.73 (for test-clean).

### Rescoring with whole lattice (G.lm_scores)

Screen Shot 2021-06-15 at 2 36 28 PM

G.lm_scores can further improve WER (5.63 --> 5.56, test-clean)

### Rescoring with n-best list (G.lm_scores, num_paths=100)
Screen Shot 2021-06-15 at 2 39 52 PM

### Rescoring with n-best list (G.lm_scores, num_paths=500)
Screen Shot 2021-06-15 at 2 40 01 PM

egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
csukuangfj: More comments are added.

danpovey:
Great!!
You can merge when you think it's ready.

PR #215, opened 2021-06-19
WIP: Add BPE training with LF-MMI.

csukuangfj:
A small vocab_size is used to avoid OOM.

@glynpu  is doing BPE CTC training. We can use his implementation once it's ready.
This pull-request is for experimental purpose.

### DIFF
--- a/egs/librispeech/asr/simple_v1/bpe_mmi_att_transformer_train.py
+++ b/egs/librispeech/asr/simple_v1/bpe_mmi_att_transformer_train.py
@@ -1,0 +1,716 @@
+#!/usr/bin/env python3
+
+# Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey
+#                                                   Haowen Qiu
+#                                                   Fangjun Kuang)
+#                2021  University of Chinese Academy of Sciences (author: Han Zhu)
+# Apache 2.0
+
+import argparse
+import logging
+import os
+import sys
+from datetime import datetime
+from pathlib import Path
+from typing import Dict, Optional
+
+import k2
+import numpy as np
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+from torch import nn
+from torch.cuda.amp import GradScaler, autocast
+from torch.nn.parallel import DistributedDataParallel as DDP
+from torch.nn.utils import clip_grad_value_
+from torch.utils.data import DataLoader
+from torch.utils.tensorboard import SummaryWriter
+
+from lhotse.utils import fix_random_seed, nullcontext
+from snowfall.common import describe, str2bool
+from snowfall.common import load_checkpoint, save_checkpoint
+from snowfall.common import save_training_info
+from snowfall.common import setup_logger
+from snowfall.data.librispeech import LibriSpeechAsrDataModule
+from snowfall.dist import cleanup_dist
+from snowfall.dist import setup_dist
+from snowfall.lexicon import Lexicon
+from snowfall.models import AcousticModel
+from snowfall.models.conformer import Conformer
+from snowfall.models.contextnet import ContextNet
+from snowfall.models.tdnn_lstm import TdnnLstm1b  # alignment model
+from snowfall.models.transformer import Noam, Transformer
+from snowfall.objectives import LFMMILoss, encode_supervisions
+from snowfall.training.diagnostics import measure_gradient_norms, optim_step_and_measure_param_change
+from snowfall.training.mmi_graph import MmiTrainingGraphCompiler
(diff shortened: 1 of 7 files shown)


egs/librispeech/asr/simple_v1/bpe_mmi_att_transformer_train.py
```diff
@@ -0,0 +1,716 @@
...
+    num_epochs = args.num_epochs
+    accum_grad = args.accum_grad
+    den_scale = args.den_scale
+    att_rate = args.att_rate
+    use_pruned_intersect = args.use_pruned_intersect
+
+    fix_random_seed(42)
+    setup_dist(rank, world_size, args.master_port)
+
+    exp_dir = Path('exp-bpe-' + model_type + '-mmi-att-sa-vgg-normlayer')
```
csukuangfj, line 490: This file is copied from mmi_att_transformer_train.py, with only this line being modified.

egs/librispeech/asr/simple_v1/generate_bpe_tokens.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+
+# Copyright (c)  2021  Xiaomi Corporation (authors: Fangjun Kuang)
+'''
+Example usage of this script:
+
+python3 ./generate_bpe_tokens.py \
+  --model-file ./data/lang_bpe/bpe_unigram_500.model > data/lang_bpe/bpe_unigram_500.tokens
```
danpovey, line 8: Is it necessary to write the model out to disk and get the tokens separately?
I'd imagine it would be very fast to train the BPE model (isn't it some very simple LM?) and get the tokens inside the same script..  this just feels a bit complicated.   But it's OK for now.

csukuangfj: > Is it necessary to write the model out to disk and get the tokens separately?

No, it's not necessary. We can generate all the needed files `tokens.txt`, `words.txt`, `lexicon.txt` in one
file. I just feel splitting them is easier to understand.

Yes, it is quite fast to train a BPE model. It's less than 1 minute for the LibriSpeech dataset.
The model is only several hundred KB and does not occupy too much disk space.

egs/librispeech/asr/simple_v1/run.sh
```diff
@@ -3,11 +3,36 @@
...
+  echo "Please set LibriSpeech dataset path before running this script"
+  exit 1
+fi
+
+echo "LibriSpeech dataset dir: $libri_dir"
+
+stage=5
+
+# settings for BPE training -- start
+vocab_size=5000
```
csukuangfj, line 33: @danpovey 

I've removed the bigram P and it is possible to do BPE LF-MMI training with vocab_size=5000.

danpovey:
BTW, the way I think we can solve the memory-blowup issue is:
 (i) use the new, more compact CTC topo
 (ii) train a bigram ARPA LM to make a compact LM, e.g. with kaldi's make_kn_lm.py; load it into k2 as P (no disambig symbols!), and remove epsilons.  k2 uses a rm-epsilon algorithm that should keep the epsilon-free LM compact, unlike OpenFst which would cause it to blow up.
  BTW, I am asking some others to add a pruning option to make_kn_lm.py.

egs/librispeech/asr/simple_v1/generate_bpe_tokens.py
csukuangfj: > Is it necessary to write the model out to disk 

According to the documentation and examples given in
https://github.com/google/sentencepiece/blob/master/python/README.md ,
it looks like the model has to be written to disk after training. There is no option to retain a memory representation
of the model without writing it to disk.

csukuangfj:
> (i) use the new, more compact CTC topo

Yes, I am using the new CTC topo.

> (ii) train a bigram ARPA LM to make a compact LM, e.g. with kaldi's make_kn_lm.py;

Will update the code to train a word piece bigram ARPA LM with `make_kn_lm.py`.

egs/librispeech/asr/simple_v1/generate_bpe_lexicon.py
```diff
@@ -0,0 +1,66 @@
...
+        Filename of words.txt.
+      excluded:
+        words in this list are not returned.
+    Returns:
+      Return a list of words.
+    '''
+    ans = []
+    with open(words_txt, 'r', encoding='latin-1') as f:
+        for line in f:
+            word, id = line.strip().split()
```
glynpu, line 33: _**id**_ is a python build-in fuction, maybe change to idx

PR #204, opened 2021-06-02
espnet-style attn_output_weight scaling and extra after-norm layer

glynpu:
Conformer structure differences are identified by loading espnet trained model into snowfall. https://github.com/k2-fsa/snowfall/pull/201
1. snowfall only scaling q; while espnet scale attn_output_weights.
2. espnet conformer has an extra layer_norm after encoder

With these two modifications and 30 epoch training, final result is a bit better(3.69 < 3.86 as reported in https://github.com/k2-fsa/snowfall/issues/154) than otherwise. 

Could you help verify their effectiveness (maybe they are just training variance)? @zhu-han @pzelasko 
BTW,  is there any mathmatics background which explains when to scaling during attn_output_weights computation? I read several papers, but failed to find a clue about this.

Rescoring WITH 4-gram lm lattice rescore
with modifications of this pr
```
avg epoch 16-20
2021-06-02 19:37:12,429 INFO [common.py:380] [test-clean] %WER 3.77% [1983 / 52576, 348 ins, 105 del, 1530 sub ]
2021-06-02 21:38:48,140 INFO [common.py:382] [test-other] %WER 7.86% [4116 / 52343, 704 ins, 260 del, 3152 sub ]
avg epoch 26-30
2021-06-02 19:25:40,616 INFO [common.py:380] [test-clean] %WER 3.69% [1938 / 52576, 386 ins, 96 del, 1456 sub ]
2021-06-02 21:45:22,304 INFO [common.py:382] [test-other] %WER 7.68% [4021 / 52343, 746 ins, 251 del, 3024 sub ]
```

results of 4-gram lattice rescore  from #154
```
avg epoch 16-20
2021-05-21 09:46:26,814 INFO [common.py:380] [test-clean] %WER 3.87% [2036 / 52576, 334 ins, 116 del, 1586 sub ]
2021-05-21 09:53:26,347 INFO [common.py:380] [test-other] %WER 8.08% [4231 / 52343, 710 ins, 241 del, 3280 sub ]
avg epoch 26-30
2021-05-22 14:53:36,527 INFO [common.py:380] [test-clean] %WER 3.86% [2030 / 52576, 345 ins, 114 del, 1571 sub ]
2021-05-22 15:00:10,075 INFO [common.py:380] [test-other] %WER 8.07% [4223 / 52343, 708 ins, 254 del, 3261 sub ]
```

Rescoring WITHOUT 4-gram lm lattice rescore
with modifications of this pr
```
avg epoch 16-20
[...]

PR #217, opened 2021-06-21
bpe ctc decoder with a released model

glynpu:
This pr release a snowfall trained model together with related decode code. 
Wer on test-clean is lower than [previously trained model](https://github.com/k2-fsa/snowfall/pull/204) with snowfall, detailed comparison as following:

avg epoch 26-30 | no rescore | no rescore | 4-gram lattice rescore | 4-gram lattice rescore
-- | -- | -- | -- | --
&nbsp; | test-clean | test-other | test-clean | test-other
[before with LF-MMI loss](https://github.com/k2-fsa/snowfall/pull/204)| 4.14 | 8.41 | 3.69 | 7.68
current | 3.97 | 9.78 | * | *

```
INFO:root:[test-clean] %WER 3.97% [2087 / 52576, 220 ins, 166 del, 1701 sub ]
INFO:root:[test-other] %WER 9.78% [5121 / 52343, 535 ins, 439 del, 4147 sub ]
```

A thing worth to mention is that： current no-rescore result(3.97 on test-clean) is got WITHOUT a 3-gram. 
Maybe the result will get lower with composing currnet ctc_topo with a 3-gram fst (I am working on this).

Another baseline of this model is an espnet released model, detailed comparison as following:
num_paths = 100 when doing n-best rescoring of row 2; 
result of row 2 is got using similar techiniques used in https://github.com/k2-fsa/snowfall/pull/201, by loading espnet released model with snowfall code.

 decoding algorithm| training tool| encoder + k2 ctc decode+no rescore | encoder + k2 ctc decode+ decoder nbest rescore | encoder + k2 ctc decode+ transformer lm nbest rescore | encoder + k2 ctc decode+ decoder nbest rescore+ transformer lm nbest rescore
-- | --|-- | -- | -- | --
decoder algorithm in espnet|espnet | * | * | * | 2.1%
k2 ctc decode in this pr |espnet| 2.97 | 2.64 | 2.43 | 2.35
k2 ctc decode in this pr| snowfall | 3.97|*|*|*|

Conclusions:
1. a better snowfall trained model is got before rescore.
2. current training pipeline is still inferior to it's espnet counterpart; if fix this, current wer 3.97% on test-clean should close to 2.97%
(related training code will be submited soon in this week; make a promise here to force me to do this quickly).

### DIFF
--- a/snowfall/models/transformer.py
+++ b/snowfall/models/transformer.py
@@ -168,7 +168,7 @@ class Transformer(AcousticModel):
             ys_out_pad = pad_list(ys_in, -1)
 
         else:
-            raise VAlueError("Invalid input for decoder self attetion")
+            raise ValueError("Invalid input for decoder self attetion")
 
 
         ys_in_pad = ys_in_pad.to(x.device)
--- a/snowfall/models/transformer.py
+++ b/snowfall/models/transformer.py
@@ -37,7 +37,7 @@ class Transformer(AcousticModel):
                  d_model: int = 256, nhead: int = 4, dim_feedforward: int = 2048,
                  num_encoder_layers: int = 12, num_decoder_layers: int = 6,
                  dropout: float = 0.1, normalize_before: bool = True,
-                 vgg_frontend: bool = False) -> None:
+                 vgg_frontend: bool = False, mmi_loss: bool = True) -> None:
         super().__init__()
         self.num_features = num_features
         self.num_classes = num_classes
@@ -64,7 +64,10 @@ class Transformer(AcousticModel):
         )
 
         if num_decoder_layers > 0:
-            self.decoder_num_class = self.num_classes + 1  # +1 for the sos/eos symbol
+            if mmi_loss:
+                self.decoder_num_class = self.num_classes + 1  # +1 for the sos/eos symbol
+            else:
+                self.decoder_num_class = self.num_classes  # bpe model already has sos/eos symbol
 
             self.decoder_embed = nn.Embedding(self.decoder_num_class, d_model)
             self.decoder_pos = PositionalEncoding(d_model, dropout)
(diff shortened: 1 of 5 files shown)


danpovey:
This was trained on 960 hours, I assume?
I'm surprised that your "k2 ctc decode" number was got *without* the 3-gram LM; I had thought you were using that.
What do you think are the differences between our trained model and the espnet-trained model?  Is it possible to compare the diagnostics from training?

glynpu:
> This was trained on 960 hours, I assume?

Yes, with full libri.
>What do you think are the differences between our trained model and the espnet-trained model?

A known huge difference is espnet use [warm_up scheduler](https://github.com/espnet/espnet/blob/a50c37b79da7cd97d86e4f475283c41685468b53/espnet2/schedulers/warmup_lr.py#L11), while I use Noam optimizer in snowfall.
with warmup_stem = 40000, and model_size=512, at each step, learning rate of espnet is around 10 times that in this exp.
So I am going to retrain to the model with changing lr-factor from 1.0 to 10.0 after current exp finished.

>Is it possible to compare the diagnostics from training?

Yes, I have reproduce espnet result and got detail training log, which will be used to diagnose my training process.

pzelasko:
You might want to check what data augmentation techniques and settings they are using and compare them with our setup. If we’re missing some techniques in Lhotse we can add them.

danpovey:
So I guess this is ready to merge?

glynpu:
Maybe @csukuangfj  is going to review this afternoon.

snowfall/models/transformer.py
```diff
@@ -146,9 +149,28 @@ def decoder_forward(self, x: Tensor, encoder_mask: Tensor, supervision: Supervis
...
+            eos_id = self.decoder_num_class - 1
+            _sos = torch.tensor([sos_id])
+            _eos = torch.tensor([eos_id])
+            ys_in = [torch.cat([_sos, torch.tensor(y)], dim=0) for y in token_ids]
+            ys_out = [torch.cat([torch.tensor(y), _eos], dim=0) for y in token_ids]
+            ys_in_pad = pad_list(ys_in, eos_id)
+            ys_out_pad = pad_list(ys_in, -1)
+
+        else:
+            raise VAlueError("Invalid input for decoder self attetion")
```
csukuangfj, line 171: VAlueError -> ValueError

csukuangfj:
+2

snowfall/models/transformer.py
glynpu: fixed.

csukuangfj:
Thanks! Merging

PR #218, opened 2021-06-21
Support to choose unigram and bigram for P in LF-MMI training.

csukuangfj:
## With unigram LM for P
```
export CUDA_VISIBLE_DEVICES="0"

./mmi_att_transformer_train.py \
  --master-port=12355 \
  --full-libri=0 \
  --use-ali-model=0 \
  --max-duration=500 \
  --use-unigram=1
```

## With bigram LM for P
```
export CUDA_VISIBLE_DEVICES="1"

./mmi_att_transformer_train.py \
  --master-port=12356 \
  --full-libri=0 \
  --use-ali-model=0 \
  --max-duration=500 \
  --use-unigram=0
```

Will report the result when it is available (Probably tomorrow morning).

### DIFF
--- a/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
+++ b/egs/librispeech/asr/simple_v1/mmi_att_transformer_decode.py
@@ -71,6 +71,7 @@ from snowfall.models.conformer import Conformer
 from snowfall.models.contextnet import ContextNet
 from snowfall.training.ctc_graph import build_ctc_topo
 from snowfall.training.mmi_graph import create_bigram_phone_lm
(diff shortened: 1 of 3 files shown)


csukuangfj:
The following shows what the unigram P and bigram P look like when there are only 3 phones: a, b, and c.

unigram
bigram

csukuangfj:
Here are the results for using unigram P and bigram P.

It shows that the WER with unigram P is worse than that of bigram P. Also, the objf of unigram P is slightly higher.

## objf values

portfolio_view

## WERs

portfolio_viewportfolio_view


portfolio_view portfolio_view

PR #221, opened 2021-06-28
Add timit recipe

luomingshuang:
The two files (timit.py and prepare.py) are for data preparation.


egs/timit/asr/simple_v1/local1/timit.py
```diff
@@ -0,0 +1,147 @@
+
```
pzelasko, line 1: I suggest to contribute this recipe to Lhotse and import it from there.

luomingshuang: OK, I will do it.

PR #222, opened 2021-06-29
Use a pre-trained bigram P for LF-MMI training

csukuangfj:
Will post the result (the WER, probably tomorrow) when it is available.

---

Here is the information about the size of P when the number of phones is 86.

|            | current P | pre-trained P | pre-trained P after epsilon removal |
|------------|-----------|---------------|-------------------------------------|
| num_states | 88        | 74            | 74                                  |
| num_arcs   | 7568      | 3634          | 7209                                |

If we are going to use word pieces with vocab_size 5000, hope that P is not going to increase its size quadratically.
Will show the size of P for word pieces soon.

### DIFF
--- a/egs/librispeech/asr/simple_v1/local/make_kn_lm.py
+++ b/egs/librispeech/asr/simple_v1/local/make_kn_lm.py
@@ -1,0 +1,377 @@
+#!/usr/bin/env python3
+
+# Copyright 2016  Johns Hopkins University (Author: Daniel Povey)
+#           2018  Ruizhe Huang
+# Apache 2.0.
+
+# This is an implementation of computing Kneser-Ney smoothed language model
+# in the same way as srilm. This is a back-off, unmodified version of
+# Kneser-Ney smoothing, which produces the same results as the following
+# command (as an example) of srilm:
+#
+# $ ngram-count -order 4 -kn-modify-counts-at-end -ukndiscount -gt1min 0 -gt2min 0 -gt3min 0 -gt4min 0 \
+# -text corpus.txt -lm lm.arpa
+#
+# The data structure is based on: kaldi/egs/wsj/s5/utils/lang/make_phone_lm.py
+# The smoothing algorithm is based on: http://www.speech.sri.com/projects/srilm/manpages/ngram-discount.7.html
+
+import sys
+import os
+import re
+import io
+import math
+import argparse
+from collections import Counter, defaultdict
+
+
+parser = argparse.ArgumentParser(description="""
+    Generate kneser-ney language model as arpa format. By default,
+    it will read the corpus from standard input, and output to standard output.
+    """)
+parser.add_argument("-ngram-order", type=int, default=4, choices=[2, 3, 4, 5, 6, 7], help="Order of n-gram")
+parser.add_argument("-text", type=str, default=None, help="Path to the corpus file")
+parser.add_argument("-lm", type=str, default=None, help="Path to output arpa file for language models")
+parser.add_argument("-verbose", type=int, default=0, choices=[0, 1, 2, 3, 4, 5], help="Verbose level")
+args = parser.parse_args()
+
+default_encoding = "latin-1"  # For encoding-agnostic scripts, we assume byte stream as input.
+                              # Need to be very careful about the use of strip() and split()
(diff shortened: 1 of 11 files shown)


csukuangfj:
Here are the WERs
(100 hours data with 10 epochs)

| test-clean        | test-other           |
| ------------- |:-------------:| 
| Screen Shot 2021-06-30 at 1 21 22 PM     | Screen Shot 2021-06-30 at 1 26 33 PM|

## Results from https://github.com/k2-fsa/snowfall/pull/212#issuecomment-861223119

122004558-25c8fb00-cde7-11eb-8e9c-2534c036ba6d

Results with this pull request are a litter worse than that of #212 

----


### Results from https://github.com/k2-fsa/snowfall/pull/218#issuecomment-865868441

 

#218 is the latest run that I have.  The experiment environments between #218 and this pull request are similar and comparable.  Compared with #218, it shows pre-trained P has a lower WER on both test-clean (5.74 vs 5.83) and test-other (15.00 vs 15.64).

danpovey:
That's interesting!
To make the graphs smaller we can consider using count cutoffs (min-counts) to take away low-count n-grams.
I'm kind of confused why there is so much difference; I would have expected the two would be very similar because
how we train the graphs is very similar to ML.   But could be due to the ARPA having smoothing, or due to differences
RE silence.

egs/librispeech/asr/simple_v1/local/make_kn_lm.py
```diff
@@ -0,0 +1,377 @@
+#!/usr/bin/env python3
+
+# Copyright 2016  Johns Hopkins University (Author: Daniel Povey)
```
csukuangfj, line 3: This file is a verbatim-copy of https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/utils/lang/make_kn_lm.py

danpovey:
I'm OK to merge this as-is.

csukuangfj:
> I'm OK to merge this as-is.

I have removed the code supporting training P on-the-fly in this pull-request.
Shall I add an option to let the user choose which kinds of P to use (This will make the code complicated.)

danpovey:
Let's keep it simple for now.

snowfall/common.py
```diff
@@ -88,9 +88,11 @@ def load_checkpoint(
             src_key = '{}.{}'.format('module', key)
             dst_state_dict[key] = src_state_dict.pop(src_key)
         assert len(src_state_dict) == 0
-        model.load_state_dict(dst_state_dict)
+        model.load_state_dict(dst_state_dict, strict=False)
```
csukuangfj, line 91: @danpovey 
Adding `strict=False` should prevent PyTorch from complaining
about extra key `P_scores` in the checkpoints.

csukuangfj:
Will merge.

danpovey:
You might want to check the strict=False option.  IIRC last time I tried it the torch code was broken and was not correctly respecting that option, and I had to make changes to torch itself locally.

csukuangfj:
@danpovey 
> You might want to check the strict=False option. IIRC last time I tried it the torch code was broken and was not correctly respecting that option, and I had to make changes to torch itself locally.

I suspect that you forgot to add it to the function `average_checkpoint()` and you added it only to `load_checkpoint()`.

I just verified it works perfectly with `strict=False` to load the old model checkpoint which has `P_scores`.

pzelasko:
After this PR, the “pretrained” P is always going to be used, right?

also did you try 3 or 4 gram (Kaldi style)? I guess it should further help

csukuangfj:
> After this PR, the “pretrained” P is always going to be used, right?

Yes, that's right. Supporting both pre-trained P and on-the-fly trained P makes the code complicated.
Pre-trained P gives a slightly better WER according to the above experiments.


---

> also did you try 3 or 4 gram (Kaldi style)? I guess it should further help

Thanks. I will try that.

danpovey:
For using a 3 or 4-gram LM, we would definitely need to do some kind of pruning, or the LM will be way too large.
Ruizhe Huang @huangruizhe is working on a self-contained Python script for Kaldi that can do that.  Let's try this after he finishes it.

Issue #220, opened 2021-06-28
CTC training speed question


yuekaizhang:
Hi, for my experiment, built-in (cudnnctc) is about 2.5 times fast than k2-ctc. I was wondering if this is normal and would like to make sure my program is correct.

I found that  `decoding_graph = k2.compose(self.ctc_topo, label_graph, treat_epsilons_specially=False)` is the reason even with `build_ctc_topo2` https://github.com/k2-fsa/snowfall/pull/209. How about considering to construct ctc loss directly from text rather than compsing the topo FST and label FST. In my experiment, it would give a very similar speed with cudnnctc.

Like the below pic:
![Screenshot2021_06_28_100754](https://user-images.githubusercontent.com/28823245/123570213-54d56880-d7fa-11eb-849d-b23966f8c1d7.jpg)

danpovey:
Thanks for doing the comparison, and sure, that's a good idea.  Yes, we should introduce a special-purpose function that constructs a batch of CTC graphs from a ragged tensor consisting of the linear symbol sequences for each one.  Perhaps @pkufool could work on that?

csukuangfj:
Shall we also consider the transition probability contained in the bigram P while constructing the graph for LF-MMI training?
(It's not an issue for CTC training.)

danpovey:
Perhaps for LF-MMI it would be best to use our current code so it takes
care of that.  But last time I checked, graph compilation does not actually
take much time since we batch things up.

Max, perhaps you could show us what code you are using for graph
compilation, e.g. are you compiling these things individually or as a batch?

yuekaizhang:
Hi, Dan, my comparison is based on the code https://github.com/k2-fsa/snowfall/blob/2dda31e14039a79b77c89bcd3bb96d52cbf60c8a/snowfall/training/ctc_graph.py#L108-L127. I used the below code to avoid composing operation. I was wondering how it works to compile them as a batch. Thanks.

    def _compile_one_and_cache_v2(self, text: torch.Tensor) -> k2.Fsa:
        text = text.tolist()
        blank_idx = 0
        num_tokens = len(text)
        S = 2 * num_tokens + 1
        final = S + 1
        arcs = []
        arcs.append([final])
        for s in range(S):
            idx = (s-1) // 2
            word_id = text[idx] if s % 2 else blank_idx
            arcs.append([s,s,word_id,word_id,0]) 
            if s > 0:
                arcs.append([s-1,s,word_id,word_id,0]) 
[...]

PR #225, opened 2021-07-01
Constructing ctc decoding graph in a batch

pkufool:
See comments #220 

> Thanks for doing the comparison, and sure, that's a good idea. Yes, we should introduce a special-purpose function that constructs a batch of CTC graphs from a ragged tensor consisting of the linear symbol sequences for each one. 

Current functions in k2 support to construct ctc decoding graph in a batch, so I think there is no other things should be done in C++ side. Fix me if I understand in a wrong way.

### DIFF
--- a/snowfall/training/ctc_graph.py
+++ b/snowfall/training/ctc_graph.py
@@ -105,23 +105,20 @@ class CtcTrainingGraphCompiler(object):
         phone_ids_with_blank = [0] + phone_ids
         self.ctc_topo = k2.arc_sort(build_ctc_topo(phone_ids_with_blank))
 
-    def compile(self, texts: Iterable[str]) -> k2.Fsa:
-        decoding_graphs = k2.create_fsa_vec(
-            [self.compile_one_and_cache(text) for text in texts])
 
-        # make sure the gradient is not accumulated
-        decoding_graphs.requires_grad_(False)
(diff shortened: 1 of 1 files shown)


danpovey:
I think we were talking about doing it in a single function, rather than a sequence of functions.
BTW, we'd have to decide which type of topology to use, in terms of how it deals with repeats of the same
symbol (i.e. do we require a blank in between?).  If we require a blank in between those repeats, the code 
becomes a little more complicated.

pkufool:
> I think we were talking about doing it in a single function, rather than a sequence of functions.
> BTW, we'd have to decide which type of topology to use, in terms of how it deals with repeats of the same
> symbol (i.e. do we require a blank in between?). If we require a blank in between those repeats, the code
> becomes a little more complicated.

pkufool:
I see,  so we will implement a function in k2 that given a lexicon fsa and ragged tensor consisting of the linear symbol sequences as input and return the ctc graph.

danpovey:
No, it would take just a sequence of phone symbols (or whatever symbols the user is using, not necessarily worsd), 
and return the linear CTC graph; I think that is his scenario.  This would not support optional silences.

pkufool:
Implements in the c++ side, see https://github.com/k2-fsa/k2/pull/776

PR #236, opened 2021-07-23
Get best matching stats from tow nbest objects

pkufool:
Functions to generate training features for _special model_ for nbest score estimation.

Require https://github.com/k2-fsa/k2/pull/786

### DIFF
--- a/snowfall/decoding/nbest_rescore.py
+++ b/snowfall/decoding/nbest_rescore.py
@@ -1,0 +1,112 @@
+# Copyright (c)  2021  Xiaomi Corporation (authors: Wei Kang)
+
+import k2
+import torch
+
+
+def get_best_matching_stats(keys: k2.Nbest, queries: k2.Nbest,
+                            max_order: int) -> torch.Tensor:
+    '''Get best matching stats on query positions.
+
+    Args:
+      keys:
+        The nbest after doing second pass rescoring.
+      queries:
+        Another nbest before doing second pass rescoring.
+      max_order:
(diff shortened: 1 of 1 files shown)


snowfall/decoding/nbest_rescore.py
```diff
@@ -0,0 +1,112 @@
...
+    keys_scores = k2.RaggedFloat(keys_tokens_shape, keys.fsa.scores.clone())
+    # scores on query positions MUST be zeros
+    queries_scores = k2.RaggedFloat(queries_tokens_shape,
+                                    torch.zeros(queries_tokens_num,
+                                                dtype=torch.float32))
+    scores = k2.ragged.cat([keys_scores, queries_scores], axis=1).values()
+
+    # we didn't remove -1 labels before
+    min_token = -1
+    eos = -1
```
pkufool, line 63: I did't remove -1 labels in nbest paths,  if I removed these labels I had to remove the scores in the corresponding positions, I did not find functions could do this thing. 

I thought the remained -1 labels should not be a problem.

The `eos` could be other value, or as a parameter just like the `max_order`.

pkufool:
@csukuangfj  I tried to generate training data from just one Nbest object, but it is too complicate, we need the init score (before second pass recoring) as one of the features,  if with just one Nbest object (after rescoring) we should pass one more parameter, say, `old_scores`.
So, I thought passing two Nbest objects to the function would be the simplest way. At that time, we have to do second pass rescoring two times, one for eath Nbest object.

danpovey:
The training time for this special is  probably going to be very small, so I wouldn't worry too much about it.

